PR #4, opened 2022-04-27
Docker env

jonfodi:
isolated environment to build firmware code across os's

### DIFF
--- a/ENV_SETUP.md
+++ b/ENV_SETUP.md
@@ -1,0 +1,18 @@
+** the script will set up the env for compiling - no need to run twice ** 
+** to test new changes, simply enter container shell and execute scons - the files will update automatically due to mounitng feature ** 
+** this also means that, once built, the build dir will be on your local machine as well. Run scons --clean in container if you want to re-build from scratch **
+
+Local changes 
+- Clone repo
+	- clone with ssh link to enable ssh connection when pulling and pushing 
+- Pull in submodules
+	- git submodule update --init --recursive 
+- Install and start docker (OS-specific)
+- Install docker-compose 
+- WINDOWS USERS: Install bash emulator: cygwin, mingw64, WSL (test optimal tool)
+- Execute compile script from dev_container dir
+	- ./buildroot.sh
+		- change perms if necessary: chmod +x buildroot.sh
+in container
+-	scons --target=<component> 
+
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,14 @@
+services:
+  buildroot:
+    image: ghcr.io/concordia-fsae/containers/ubuntu-jammy-2022.04.21
+    container_name: buildroot
+    hostname: builder
+    volumes: # mount pts
+      - .:/firmware
+    user: builduser
+    tty: true # open terminal for interacting
+    stdin_open: true
+    privileged: true
+    devices:
+      # map host's usb buses to container's buses for st-link programmer
+      - "/dev/bus/usb/:/dev/bus/usb/"
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+**
+!requirements.txt
--- a/buildroot.sh
+++ b/buildroot.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+docker-compose pull
+docker-compose run --rm buildroot
+
--- a/components/steering_wheel/SConscript
+++ b/components/steering_wheel/SConscript
@@ -255,5 +255,6 @@ env.Alias(
         "build/steering_wheel.elf",
         '-ex "target extended-remote localhost:3333"',
         '-ex "monitor reset"',
+        '-ex load', # fixed an issue found when debugging IO global struct for aggregated Digital and Analog IO
     ),
 )
--- a/site_scons/site_tools/arm-none-eabi.py
+++ b/site_scons/site_tools/arm-none-eabi.py
@@ -1,7 +1,7 @@
 from os.path import splitext
 from SCons.Script import *
 
-TOOL_PATH = Dir("#/embedded/toolchains/gcc-arm-none-eabi/bin/")
+TOOL_PATH = Dir("/opt/toolchains/gcc-arm-none-eabi/bin/")
 
 
 def _program(target, source, env, for_signature):


dev_container/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+# download compiler files 
+RUN wget https://developer.arm.com/-/media/Files/downloads/gnu/11.2-2022.02/binrel/gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz
+
+# extract tar and remove .xz 
+RUN  tar -xf /gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz \
+	&& rm -f /gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz
+
+# define default dir when entering shell to be dir of project 
+WORKDIR /usr/project/
```
rickyelopez, line 16: `/firmware`

dev_container/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN apt-get -qq update \
+        && apt-get -y install scons python3.5 wget vim 
+
+RUN apt install python3-pip -y
+
+# download compiler files 
+RUN wget https://developer.arm.com/-/media/Files/downloads/gnu/11.2-2022.02/binrel/gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz
+
+# extract tar and remove .xz 
+RUN  tar -xf /gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz \
```
rickyelopez, line 12: download in the shell script instead

dev_container/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:jammy-20220421
+
+RUN apt-get -qq update \
+        && apt-get -y install scons python3.5 wget vim 
+
+RUN apt install python3-pip -y
```
rickyelopez, line 6: combine with previous pip commands

dev_container/compile_docker.sh
```diff
@@ -0,0 +1,47 @@
...
+	 
+	# pull base image 
+	docker pull ${BASE_IMAGE}
+	
+	# build base ubuntu image  
+	docker build -t ${IMAGE_NAME} .
+fi
+
+# start and mount program files to container workdir 
+docker run -t -d -v ${C_CODE}:${WORKDIR} --name ${CONTAINER_NAME} ${IMAGE_NAME}
```
rickyelopez, line 34: user a `docker-compose.yml` instead

dev_container/compile_docker.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# install requirements 
+docker exec ${CONTAINER_NAME} pip3 install -r ${WORKDIR}/requirements.txt
```
rickyelopez, line 37: move to Dockerfile

dev_container/compile_docker.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# rename and move compiler files to appropriate dir 
+docker exec ${CONTAINER_NAME} mv /gcc-arm-11.2-2022.02-x86_64-arm-none-eabi/ /usr/project/embedded/platforms/toolchain-gccarmnoneeabi/
```
rickyelopez, line 40: download the compiler in this script instead of in the Dockerfile, then you can extract here

also use `tar -xf <archive> -C embedded/toolchains/gcc-arm-none-eabi`

dev_container/compile_docker.sh
```diff
@@ -0,0 +1,47 @@
...
+
+# open containers shell
+docker exec -it ${CONTAINER_NAME} bash
```
rickyelopez, line 43: this will become
`docker-compose up`

rickyelopez:
move `compile_docker.sh` and `Dockerfile` to the root of the repo, add `docker-compose.yml`

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:jammy-20220421
+
+#ENV REFRESHED_AT 2022-04-28 # clear cache
+
+RUN apt-get -qq update \
+        && apt-get -y install scons python3.10 wget python3-pip
+
+COPY requirements.txt /firmware/
```
rickyelopez, line 8: don't put this in `/firmware` or else it will conflict when you mount the volume later. Just put it in `/requirements.txt`

ENV_SETUP.md
```diff
@@ -0,0 +1,16 @@
+** the script will set up the env for compiling - no need to run twice ** 
+** to test new changes, simply enter container shell and execute scons - the files will update automatically due to mounitng feature ** 
+** this also means that, once built, the build dir will be on your local machine as well. Delete if you want to re-build from scratch **
```
rickyelopez, line 3: that's what `--clean` is for, scons will handle that for you

ENV_SETUP.md
```diff
@@ -0,0 +1,16 @@
...
+
+Local changes 
+- Clone repo
+	- clone with ssh link to enable ssh connection when pulling and pushing 
+- Pull in submodules
+	- git submodule --update --init recursive 
+- Install and start docker (OS-specific)
+- Execute compile script from dev_container dir
+	- ./compile_docker.sh
```
rickyelopez, line 12: hmmmm I'm just now realizing this won't work nicely on windows unless you're using WSL2. Probably fine to proceed but might be worth noting

compile_docker.sh
```diff
@@ -0,0 +1,51 @@
...
+	
+# build base ubuntu image  
+docker build -t ${IMAGE_NAME} .
+
+docker-compose up -d
+
+# download compiler 
+
+# search recursively through repo dir for compiler 
+find . -name "gcc-arm-11.2-2022.02-x86_64-arm-none-eabi*" > compiler_exist.txt
```
rickyelopez, line 25: why? The compiler has to exist in one specific place for this whole thing to work, no need to check elsewhere

compile_docker.sh
```diff
@@ -0,0 +1,51 @@
...
+
+if [ -s compiler_exist.txt ]; then
```
rickyelopez, line 27: don't create a text file, just use a variable in this script

compile_docker.sh
```diff
@@ -0,0 +1,51 @@
...
+
+	echo "Compiler files present. Skipping download"
+else
+	echo "No compiler files found. Downloading."
+	docker exec ${CONTAINER_NAME} wget https://developer.arm.com/-/media/Files/downloads/gnu/11.2-2022.02/binrel/gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz
```
rickyelopez, line 32: why are you doing this in the context of the container? Just execute the wget command from the shell script, no need to involve the docker container

compile_docker.sh
```diff
@@ -0,0 +1,51 @@
...
+        
+        # TODO: update path to match latest changes 
+        #extract and move to appropriate dir 
+        echo "Extracting tar file..."
+	docker exec ${CONTAINER_NAME} tar -xf gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz -C embedded/platforms/
```
rickyelopez, line 37: same comment as above here, just execute the command directly without docker. Also I think the destination path is wrong, should be `embedded/toolchains/gcc-arm-none-eabi` once extracted

compile_docker.sh
```diff
@@ -0,0 +1,51 @@
...
+
+        # rename compiler files and remove .xz 
+        docker exec ${CONTAINER_NAME} mv embedded/platforms/gcc-arm-11.2-2022.02-x86_64-arm-none-eabi ${COMPILER_DIR} && rm -f gcc-arm-11.2-2022.02-x86_64-arm-none-eabi.tar.xz
```
rickyelopez, line 40: Pretty sure you can specify the destination dir when unpacking with `tar` above.
Also, same comment as before re: not executing with docker

docker-compose.yml
```diff
@@ -0,0 +1,8 @@
+services:
```
rickyelopez, line 1: nice

ENV_SETUP.md
jonfodi: ya was worried about this too for windows users. seems like theres no way of running bash without WSL. should we add it as a requirement?

compile_docker.sh
jonfodi: I was thinking of windows users or linux/mac users that dont have wget binary - since the container does I just ran the cmd in the container. if were making it a requirement to have WSL we can also say install wget

compile_docker.sh
jonfodi: im using the path from the previous file structure. ill rebase then implement this change

compile_docker.sh
rickyelopez: that's a good idea but I don't think this shell script will run on windows regardless, only on linux/mac. Might be a good idea to test with cygwin on windows at some point.

ENV_SETUP.md
rickyelopez: you could also do it in a cygwin or mingw64 shell I think, in which case my previous comment might be wrong.

Dockerfile
```diff
@@ -9,6 +9,9 @@ COPY requirements.txt /
 
 RUN pip3 install -r /requirements.txt
 
+# Get ST-Link drivers
+RUN apt-get install stlink-tools
```
rickyelopez, line 13: Why is this tool needed? I already added support for flashing and debugging the chip using openOCD. Lmk if you want me to show you how

Eltom777: We were not able to flash the chip within the container using the St-link programmer. As a result, we add the drivers this way. Are you free today ? We could discuss it.

rickyelopez: this isn't necessary.
`scons --target=stw upload` to flash using openOCD
`scons --target=stw openocd-gdb` to debug using openOCD and GDB
The line you added to the docker-compose might still be necessary though. Can you test the above and let me know if that works? We will have to add openOCD to this repo at some point, like how we did for gcc

Eltom777: We ran the command `scons--target=stw upload`  inside the container. It failed. To fix this issue, we added the stlink-tools to the container. Now the stw component can be successfully flashed to the device.

I suspect openocd-gdb will also cause issues since openocd is not installed in the container.

rickyelopez: Ahhhh shit yeah I just remembered that I _did_ use stlink tools to implement the upload command. Sorry my bad, I got confused. Yeah this looks fine, and yeah like you said we'll have to install openocd in the container as well. Fortunately that shouldn't be very difficult, can probably just `apt-get install openocd` in the dockerfile, but would be good to pin the versions of both the st-link tools as well as openocd. Can you guys work on that?

Eltom777: Should we load openocd's stm32f103 config file in the script as default or should we let the user do that ?

rickyelopez: The tool I wrote (`site_scons/site_tools/openocd.py`) already allows the consumer to specify which config file to use. The `stw` SConscript file passes the correct config file to openocd

Dockerfile
```diff
@@ -9,8 +9,11 @@ COPY requirements.txt /
 
 RUN pip3 install -r /requirements.txt
 
-# Get ST-Link drivers
-RUN apt-get install stlink-tools
+# Get ST-Link drivers, openocd and gdb
+RUN apt-get update && apt-get -y install \ 
+    stlink-tools \
```
rickyelopez, line 14: Can you please pin (i.e. install specific versions) the versions of all these? It's very important to have a deterministic setup. With the way this works right now, the docker env could be completely different from one day to the next if the packages get updated

compile_docker.sh
```diff
@@ -0,0 +1,55 @@
...
+	
+	#extract and move to appropriate dir 
+        echo "Extracting tar file..."
+	mkdir embedded/toolchains/gcc-arm-none-eabi
+	docker exec ${CONTAINER_NAME} tar -xf ${COMPILER_TAR} -C embedded/toolchains/gcc-arm-none-eabi/ --strip-components 1
+
+fi
+
+# open containers shell
+docker exec -it ${CONTAINER_NAME} bash
```
rickyelopez, line 51: add `--rm` to this command (after `-it`) so that you don't get a new container every time you run it

rickyelopez:
- Refactored this to work somewhat differently.
- Dockerfile moved to containers repo
- Moved the compiler download to the Dockerfile
- built the docker image and pushed it up to the GHCR [see here](https://github.com/concordia-fsae/containers/pkgs/container/containers%2Fubuntu-jammy-2022.04.21)
- Updated docker-compose to use this image

Now all the script has to do is launch the container. lmk what you guys think

Dockerfile
JoshLafleur: Verified resolved in https://github.com/concordia-fsae/containers

ENV_SETUP.md
```diff
@@ -0,0 +1,18 @@
...
+Local changes 
+- Clone repo
+	- clone with ssh link to enable ssh connection when pulling and pushing 
+- Pull in submodules
+	- git submodule update --init --recursive 
+- Install and start docker (OS-specific)
+- Install docker-compose 
+- WINDOWS USERS: Install bash emulator: cygwin, mingw64, WSL (test optimal tool)
+- Execute compile script from dev_container dir
+	- ./compile_docker.sh
```
JoshLafleur, line 14: New script under buildroot.sh

ENV_SETUP.md
```diff
@@ -0,0 +1,18 @@
...
+		- change perms if necessary: chmod +x compile_docker.sh
```
JoshLafleur, line 15: same as above

JoshLafleur (changes requested):

JoshLafleur (approved):

rickyelopez:
let's get these commits cleaned up and then I'm happy to merge this. Anybody want to do that? I don't mind doing it but would rather teach one of you to do it

rickyelopez (approved):
Looks good to me!

rickyelopez:
@jonfodi can you review to make sure we didn't break anything?

jonfodi:
> @jonfodi can you review to make sure we didn't break anything?

Looks good to me @rickyelopez - i can compile all the stw code with the new structure/framework. I dont have an mc/st-link with me right now but im assuming Tom and Josh tested the flashing and debugging features with the new framework as well.

JoshLafleur:
@jonfodi @rickyelopez yes confirmed both upload and openocd-gdb is working. I pushed an edit to the sconscript file which adds the execution of load to the gdb command as i found i was having issues with some symbols when working with the io_digital branch that load fixed. Also noticed that openocd gdb documentation also has this command included

rickyelopez:
Nice, ok go ahead and approve then, and let's merge this

rickyelopez:
I'm just going to merge this.

JoshLafleur:
Agreed

PR #9, opened 2022-06-01
Io digital

JoshLafleur:
Untested EXTI implementation of IO Digital


.gitignore
```diff
@@ -273,3 +273,8 @@ tags
 .ionide
 
 # End of https://www.toptal.com/developers/gitignore/api/vim,c,c++,python,visualstudiocode,platformio
+
+.env
+.docker
+settings.json
+.gitignore
```
rickyelopez, line 280: remove `.gitignore`

components/steering_wheel/include/IO_Analog.h
```diff
@@ -29,11 +29,11 @@ typedef struct
 
     uint32_t test;
 
-} IO_S;
+} IO_ANALOG_S;
```
rickyelopez, line 32: `IO_Analog_S`

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+ * Defines
+ * *********************************************/
+
+#define NUM_INPUTS 7
+
+/************************************************
+ * Typedefs
+ * *********************************************/
+
+enum INPUT
```
rickyelopez, line 22: don't do this, the only reason you'd really do this is if you're iterating over them

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+};
+// static_assert(COUNT_INPUT == NUM_INPUTS);
+
+#if 8 >= NUM_INPUTS
+typedef unsigned char io_digital_t;
+#elif 16 >= NUM_INPUTS
+typedef unsigned short io_digital_t;
+#else
+typedef unsigned int io_digital_t;
+#endif
```
rickyelopez, line 41: don't do this

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+
+typedef struct
+{
+    io_digital_t input;
```
rickyelopez, line 45: instead of the `io_digital_t` add one of these inside the `io_digital_S` struct
```
    struct
    {
        bool switch1:1;
        bool switch2:1;
        etc.
    } bit;
```
it's called a bitfield, compiler can optimize everything into a minimal amount of memory

components/steering_wheel/HW/HW_gpio.c
```diff
@@ -50,9 +50,9 @@ void MX_GPIO_Init(void)
 
     // Configure GPIOB input pins
     GPIO_InitStruct.Pin = SL_DATA_Pin | FT_PDN_Pin |
-                          BTN_MIDDLE_Pin | BTN_TOP_LEFT_Pin | BTN_TOP_RIGHT_Pin |
                           ROT_ENC_1_A_Pin | ROT_ENC_1_B_Pin |
                           ROT_ENC_2_A_Pin | ROT_ENC_2_B_Pin |
+                          BTN_MIDDLE_Pin | BTN_TOP_LEFT_Pin | BTN_TOP_RIGHT_Pin |
```
rickyelopez, line 55: can you revert this please? fewer diffs is always better

.gitignore
```diff
@@ -273,3 +273,7 @@ tags
 .ionide
 
 # End of https://www.toptal.com/developers/gitignore/api/vim,c,c++,python,visualstudiocode,platformio
+
+.env
```
rickyelopez, line 277: Please put any custom additions to this file at the top before the autogenerated section (with the other stuff I added) rather than at the bottom

components/steering_wheel/HW/HW_intc.c
```diff
@@ -9,9 +9,9 @@
  *                             I N C L U D E S
  ******************************************************************************/
 
-#include "stm32f1xx.h"
 #include "HW_intc.h"
-
+#include "SystemConfig.h"
```
rickyelopez, line 13: why did you make these changes?

JoshLafleur: This was for when it was interrupt driven, I forgot to remove, have just done it.

components/steering_wheel/include/Display/CommonDisplay.h
```diff
@@ -87,7 +87,7 @@ static void common_display(void)
 
 
     char tempMCU[10] = { 0U };
-    snprintf(tempMCU, 10, "% 2.*f", 2, IO.temp.mcu);
+    snprintf(tempMCU, 10, "% 2.*f", 2, IO_ANALOG.temp.mcu);
```
rickyelopez, line 90: Actually, now that I'm thinking about this more, maybe we should make a common `IO.h` file which contains a single `IO` struct with `analog` and `digital` structs as members... I feel like that will be less annoying to use than having to reference the structs separately

components/steering_wheel/include/IO_Analog.h
```diff
@@ -1,6 +1,6 @@
 /**
- * IO.h
- * IO Module Header
+ * IO_Analog.h
+ * IO Analog Module Header
```
rickyelopez, line 3: nit: `Analog IO Module Header`

components/steering_wheel/include/Display/CommonDisplay.h
rickyelopez: no nevermind this is a bad idea

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,36 @@
+/**
+ *IO_Digital.h
```
rickyelopez, line 2: nit: space after the *

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,36 @@
...
+ *Digital IO Module Header
+ */
+
+#pragma once
+
+/******************************************************************************
+ *                             I N C L U D E S
+ ******************************************************************************/
+
+#include <stdbool.h>
```
rickyelopez, line 12: just include `Types.h`

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,36 @@
...
+    {
+        bool switch1:1;
+        bool switch2:1;
+        bool switch4:1;
+        bool switch5:1;
+        bool button_top_left:1;
+        bool button_middle:1;
+        bool button_top_right:1;
+    } input;
+} io_digital_S;
```
rickyelopez, line 30: the global struct should be all-caps, the local one in the C file should be lowercase. so this should be `IO_Digital_S`

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,36 @@
...
+
+/******************************************************************************
+ *                           P U B L I C  V A R S
+ ******************************************************************************/
+
+extern io_digital_S io_digital;
```
rickyelopez, line 36: same comment as above, should be `IO_DIGITAL`

components/steering_wheel/src/IO_Analog.c
```diff
@@ -1,6 +1,6 @@
 /**
- * IO.c
- * The IO module controls interactions with the chip IO
+ * IO_ANALOG.c
```
rickyelopez, line 2: nit: `IO_Analog.c`

components/steering_wheel/src/IO_Analog.c
```diff
@@ -1,6 +1,6 @@
...
+ * The IO Analog module controls interactions with the chip analog IO
```
rickyelopez, line 3: nit: `The Analog IO module... with the chip's analog IO`

components/steering_wheel/src/IO_Digital.c
```diff
@@ -0,0 +1,64 @@
...
+// Other includes
+#include "ModuleDesc.h"
+#include "Utility.h"
+#include "SystemConfig.h"
+
+/******************************************************************************
+ *                           P U B L I C  V A R S
+ ******************************************************************************/
+
+io_digital_S io_digital;
```
rickyelopez, line 25: Global var should be uppercase

components/steering_wheel/src/IO_Digital.c
```diff
@@ -0,0 +1,64 @@
...
+{
+    memset(&io_digital, 0, sizeof(io_digital_S));
+}
+
+/**
+ * @brief Periodic function checking input states
+ */
+static void IO_Digital_10Hz_PRD(void)
+{
+    //Because of pull-up, signal is active low
```
rickyelopez, line 44: nit: space after `//`

components/steering_wheel/src/IO_Digital.c
```diff
@@ -0,0 +1,64 @@
...
+    io_digital.input.switch1 = (HAL_GPIO_ReadPin(SW1_GPIO_Port, SW1_Pin) == GPIO_PIN_RESET)
```
rickyelopez: while this does work, I was thinking of a more generic approach.

Make a struct type which has members for the GPIO port and pin, as well as a pointer to the destination (which you won't be able to do right now because you can't make a pointer to a member of a bitfield. However, I'm working on a FLAGS api which you can use here instead)

Then make a static array containing those structs, where each entry refers to one of the IO pins. You can make one of these arrays for each refresh rate we want to sample at (just one is fine for now). Then in each of the periodic functions, just iterate over the relevant array and do a read on that port/pin and populate the target variable.

components/steering_wheel/HW/HW_intc.c
```diff
@@ -12,7 +12,6 @@
 #include "stm32f1xx.h"
 #include "HW_intc.h"
 
-
```
rickyelopez, line 15: please just completely revert the changes to this file, it shouldn't even be included in the pull request.

.gitignore
```diff
@@ -272,4 +276,4 @@ tags
 .history
 .ionide
 
-# End of https://www.toptal.com/developers/gitignore/api/vim,c,c++,python,visualstudiocode,platformio
+# End of https://www.toptal.com/developers/gitignore/api/vim,c,c++,python,visualstudiocode,platformio
```
rickyelopez, line 279: nit: add a newline at the end to get rid of this diff

components/steering_wheel/include/IO_Digital.h
```diff
@@ -0,0 +1,36 @@
...
+ * Digital IO Module Header
+ */
+
+#pragma once
+
+/******************************************************************************
+ *                             I N C L U D E S
+ ******************************************************************************/
+
+#include "Types.h"
```
rickyelopez, line 12: nit: one newline after the header, two newlines before the next header

components/steering_wheel/src/IO_Digital.c
```diff
@@ -0,0 +1,64 @@
...
+/******************************************************************************
+ *                       P U B L I C  F U N C T I O N S
+ ******************************************************************************/
+
+/*
+ * @brief Initializes the Digital IO Init
+ */
+static void IO_Digital_init(void)
+{
+    memset(&io_digital, 0, sizeof(IO_Digital_S));
```
rickyelopez, line 36: second parameter should be `0x00` or `0U` (I prefer `0x00`) for clarity and to make sure the compiler does what it should

JoshLafleur:
Tested confirmed that the IO_Digital module is working through openocd-gdb as of commit 93a4e25

components/steering_wheel/include/IO/IO.h
```diff
@@ -0,0 +1,37 @@
...
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef struct
+{
+  IO_Analog_S *Analog;
+  IO_Digital_S *Digital;
+} IO_S;
```
rickyelopez, line 30: Why? I don't think this is necessary

components/steering_wheel/src/IO/IO.c
```diff
@@ -0,0 +1,38 @@
+/**
```
rickyelopez, line 1: A whole C file just to make a global? Not worth it, unless there's a benefit I'm not seeing here

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+#define BTN_TOP_RIGHT_FLAG  0x40
+#define NUM_INPUTS          7U /**< Must be equal to the count of flags */
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef 
+#if 8 >= NUM_INPUTS
+unsigned char
```
rickyelopez, line 37: you should never use these types in embedded programming. Always use types that specify the size, i.e. `uint8_t`, `uint16_t`, etc.

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+#define BTN_MIDDLE_FLAG     0x20
+#define BTN_TOP_RIGHT_FLAG  0x40
+#define NUM_INPUTS          7U /**< Must be equal to the count of flags */
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef 
+#if 8 >= NUM_INPUTS
```
rickyelopez, line 36: Why? What's the point of this? What is the benefit of this vs. just specifically choosing the type and adding a static assert to ensure that the pins will fit in that type?

components/steering_wheel/src/IO/IO.c
JoshLafleur: Has to do with a previous comment you made about making a global struct that has all of the IO data aggregated. It was either splitting IO_Digital, IO_Analog, and having and aggregate IO or mashing everything into a single IO, and I thought that this method would be cleaner. What do you think about the options?

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: More scalable/reusable from what I think, to transfer this module to another system would only take changing the defined flags and NUM_INPUTS of the IO_Digital.h file. Less to adapt long term

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: resolved very good point

components/steering_wheel/include/IO/IO_Digital.h
rickyelopez: but how is that more portable than just making a variable of the correct size by hand?

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+#elif 64 >= NUM_INPUTS
+uint64_t
+#endif
+IO_Digital_Flag_t; /**< Current Implementation can handle upto 64 inputs */
+
+typedef struct 
+{
+    GPIO_TypeDef*       gpiox;
+    uint16_t            pin;
+    IO_Digital_Flag_t   flag;
```
rickyelopez, line 51: with this strategy, if we have 64 inputs, this variable will be 64 bits long and the array that contains 64 of this struct will also contain 64x 64bit variables. This is a huge amount of memory use.

components/steering_wheel/src/IO/IO.c
rickyelopez: yeah but then I commented that it was a bad idea lol. What's the benefit of this? `IO->Analog` vs `IO_Analog`. Second one has one less character and takes up less memory

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: Very strong point I had not realized, I have returned to enum type due to iterability and removed any need to store the flag

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,55 @@
...
+
+#include "stm32f103xb.h" /**< Needed for GPIO_TypeDef */
+#include "Types.h" 
+
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef enum {
```
rickyelopez, line 22: newline before `{`

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,55 @@
...
+    SW1 = 0U,
+    SW2,
+    SW4,
+    SW5,
+    BTN_TOP_LEFT,
+    BTN_MIDDLE,
+    BTN_TOP_RIGHT,
+    NUM_INPUTS
+} Input_Source;
```
rickyelopez, line 31: `digitalInput_E`

components/steering_wheel/src/IO/IO.c
JoshLafleur: Ahh I didnt remember the second comment. In that case I will return it to IO_A and IO_D as it was

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: It would calculate automatically based on enum size so could capitalize on that note. Will edit to directly setting it manually

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -53,7 +53,7 @@ static void IO_Digital_init(void)
 static void IO_Digital_10Hz_PRD(void)
 {
     /**< Because of pull-up, signal is active low */
-    for (Input_Source i = 0; i < NUM_INPUTS; i++)
+    for (digitalInput_E i = 0; i < NUM_INPUTS; i++)
```
rickyelopez, line 56: nit: technically it should be `(digitalInput_E)0U` to avoid the implicit cast

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -28,28 +32,27 @@ typedef enum {
...
 } IO_Digital_Input_S;
 
 typedef struct
 {
-    const IO_Digital_Input_S    inputs[NUM_INPUTS];
-    IO_Digital_Flag_t           input_state;
-} IO_Digital_S;
\ No newline at end of file
+    const IO_Digital_Input_S inputs[NUM_INPUTS];
+    IO_Digital_Flag_t        input_state;
```
rickyelopez, line 50: please use camel case for member variables. `inputState`

rickyelopez:
why did you merge docker_env into this? you shouldn't do that

JoshLafleur:
> why did you merge docker_env into this? you shouldn't do that

As docker_env is the base of the pull request, I updated the most recent changes there into io_digital

JoshLafleur:
okay I will revert that and rebase. Also looking into squashing now for the first time and will move ahead with that to tidy it up

rickyelopez:
ok let me know if you need help, it's pretty simple though just
`git rebase -i HEAD~10` where 10 is the number of commits (including the top commit) that you want to edit. then you just change the word `pick` to either `squash` or `fixup` to combine commits (squash if you want to edit the message, fixup just to take the message from the older commit)

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -0,0 +1,73 @@
...
+/******************************************************************************
+ *                       P U B L I C  F U N C T I O N S
+ ******************************************************************************/
+
+/**
+ * @brief Initializes the Digital IO Init
+ */
+static void IO_Digital_init(void)
+{
+    memset(&IO_DIGITAL.inputState, 0x00, sizeof(IO_Digital_S) - sizeof(IO_Digital_Input_S)*NUM_INPUTS);
```
rickyelopez, line 48: set the whole struct equal to 0, not just some of the contents.
`memset(&IO_DIGITAL, 0x00, sizeof(IO_DIGITAL));`

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @brief Periodic 10Hz function checking input states
+ */
+static void IO_Digital_10Hz_PRD(void)
+{
+    /**< Because of pull-up, signal is active low */
+    for (digitalInput_E i = (digitalInput_E) 0U; i < NUM_INPUTS; i++)
+    {
+        if (HAL_GPIO_ReadPin(IO_DIGITAL.inputs[i].gpiox, IO_DIGITAL.inputs[i].pin) == GPIO_PIN_RESET)
```
rickyelopez, line 59: perfect example of somewhere where it makes more sense to implement the function ourselves rather than use the HAL function so that we can customize it to our needs in the future. Here's what this function does in the HAL:
```
GPIO_PinState HAL_GPIO_ReadPin(GPIO_TypeDef *GPIOx, uint16_t GPIO_Pin)
{
  GPIO_PinState bitstatus;

  /* Check the parameters */
  assert_param(IS_GPIO_PIN(GPIO_Pin));

  if ((GPIOx->IDR & GPIO_Pin) != (uint32_t)GPIO_PIN_RESET)
  {
    bitstatus = GPIO_PIN_SET;
  }
  else
  {
    bitstatus = GPIO_PIN_RESET;
  }
  return bitstatus;
}
```

components/steering_wheel/SConscript
```diff
@@ -140,7 +140,8 @@ project_source_files = [
     PROJECT_SOURCE_DIR.File("SystemManager.c"),
     PROJECT_SOURCE_DIR.File("Module.c"),
     PROJECT_SOURCE_DIR.File("CAN/CANIO-tx.c"),
-    PROJECT_SOURCE_DIR.File("IO.c"),
+    PROJECT_SOURCE_DIR.File("IO/IO_Analog.c"),
```
rickyelopez, line 143: we can do this in a future PR, but I think we should actually go with `ADC` for the analog stuff and just `IO` for the digital side.

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+#pragma once
+
+/******************************************************************************
+ *                             I N C L U D E S
+ ******************************************************************************/
+
+#include "Types.h"
+#include "stm32f103xb.h" /**< Needed for GPIO_TypeDef */
```
rickyelopez, line 15: should include `gpio.h` instead. It'll do the same thing at the end of the day, but allows for further abstraction for other chips in the future.

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+
+// System Includes
+#include <assert.h>
```
rickyelopez, line 18: this should not be necessary. Look at the static asserts I did in IO_analog.c.

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+    SW2,
+    SW4,
+    SW5,
+    BTN_TOP_LEFT,
+    BTN_MIDDLE,
+    BTN_TOP_RIGHT,
+    NUM_INPUTS
+} digitalInput_E;
+
+typedef uint8_t IO_Digital_Flag_t; /**< Current Implementation can handle upto 64 inputs */
```
rickyelopez, line 37: Is this comment accurate? Seems like a uint8 could only handle 8 inputs?

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+static_assert(NUM_INPUTS <= (sizeof(IO_Digital_Flag_t) * 8), 
+    "Size of DigitalInput_E must be less than or equal to the width of IO_Digital_Flag_t");
+
+typedef struct
+{
+    GPIO_TypeDef* gpiox;
```
rickyelopez, line 43: `port` instead of `gpiox` please

components/steering_wheel/src/IO/IO_Digital.c
JoshLafleur: Should it be standard throughout the whole project to interact with the hardware through registers (not considering initialisation type activities)? All my historical usage has been through HAL more or less so would be new to me but is faisable. When you say customize to our needs, could you give me a use case so I can better understand?

components/steering_wheel/src/IO/IO_Digital.c
JoshLafleur: I had this included at one point but I do not remember now the point of it and I forgot to remove will. Will make the edit tomorrow

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,58 @@
...
+    uint16_t      pin;
+} IO_Digital_Input_S;
+
+typedef struct
+{
+    const IO_Digital_Input_S inputs[NUM_INPUTS];
```
rickyelopez, line 49: This shouldn't be part of the global struct, it should be internal to the module.

components/steering_wheel/SConscript
JoshLafleur: So change the IO_Analog over to ADC.c? Just to confirm proper understanding

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -0,0 +1,73 @@
...
+#include "Utility.h"
+#include "SystemConfig.h"
+
+
+/******************************************************************************
+ *                           P U B L I C  V A R S
+ ******************************************************************************/
+
+IO_Digital_S IO_DIGITAL = {
+    .inputs[SW1] = { .gpiox = SW1_GPIO_Port, .pin = SW1_Pin },
```
rickyelopez, line 29: please use the format
```
.inputs = {
    [SW1] = { ... },
    [SW2] = { ... },
    etc.
}
```

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: No I had not updated when making the edit. Will fix this thanks for pointing out

components/steering_wheel/SConscript
rickyelopez: it's ok you don't have to do it here, we can do that another time. Yeah the filenames as well as struct names and stuff

components/steering_wheel/src/IO/IO_Digital.c
rickyelopez: see my other comment for example of what could be useful. Yeah the idea would be that we have our own API for accessing GPIO stuff, and we can implement it in different ways for each chip, but that will come down to the chip initialization stuff. Like what I did for the CAN layer, we just call CanSend for example, and the low level stuff gets done directly in can.c

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: I looked through the CAN layer and I am having a bit of a hard time understanding really what the abstraction would be for this. I understand the benefit of abstracting away from GPIO_TypeDef, however when I try to think of the actual applicationg of it I only see currently
typedef GPIO_TypeDef IO_T;
and having the HAL_GPIO_ReadPin being abstracted though a function GPIO_ReadPin which would just call HAL_GPIO_ReadPin. Is this the idea you had in mind to create that abstraction?

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -0,0 +1,73 @@
...
+#include "ModuleDesc.h"
+#include "SystemConfig.h"
+#include "Utility.h"
+
+
+/******************************************************************************
+ *                         P R I V A T E  V A R S
+ ******************************************************************************/
+
+const HW_GPIO_Input_S inputs[NUM_INPUTS] = {
```
rickyelopez, line 28: nit: formatting. With a table like this it's always nice to have the columns aligned so everything is clear

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -0,0 +1,73 @@
...
+ * @brief Periodic 10Hz function checking input states
+ */
+static void IO_Digital_10Hz_PRD(void)
+{
+    /**< Because of pull-up, signal is active low */
+    for (digitalInput_E i = (digitalInput_E)0U; i < NUM_INPUTS; i++)
+    {
+        if (HW_GPIO_ReadPin(&inputs[i]) == HW_PIN_RESET)
+        {
+            IO_DIGITAL.inputState |= 0x01 << (unsigned int)i;
```
rickyelopez, line 61: nit: don't use unsigned int. What size data is this? use `uint8_t` or different size.

components/steering_wheel/src/IO/IO_Digital.c
```diff
@@ -0,0 +1,73 @@
...
+{
+    memset(&IO_DIGITAL, 0x00, sizeof(IO_Digital_S));
+}
+
+/**
+ * @brief Periodic 10Hz function checking input states
+ */
+static void IO_Digital_10Hz_PRD(void)
+{
+    /**< Because of pull-up, signal is active low */
```
rickyelopez, line 56: whether it's `pull up`, `pull down`, or `none` should be specified in the gpio struct, and then it should be checked here and compared accordingly. Much more expandable that way

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,48 @@
...
+
+typedef enum
+{
+    SW1 = 0U,
+    SW2,
+    SW4,
+    SW5,
+    BTN_TOP_LEFT,
+    BTN_MIDDLE,
+    BTN_TOP_RIGHT,
```
rickyelopez, line 30: seems like the paddle switches are missing?

components/steering_wheel/HW/HW_gpio.c
```diff
@@ -58,3 +58,26 @@ void MX_GPIO_Init(void)
...
+ * @brief Reads the input pin state
+ * @param[in] input Input pin
+ * @return HW pin state
+ */
+HW_GPIO_PinState_E HW_GPIO_ReadPin(const HW_GPIO_Input_S *input)
+{
+    GPIO_PinState bitstatus;
+
+    /* Check the parameters */
+    assert_param(IS_GPIO_PIN(input->pin));
```
rickyelopez, line 72: we don't need this. This is useful if you're an API maintainer where the API will be used by external folks, but that's not the case here.

components/steering_wheel/HW/include/HW_gpio.h
```diff
@@ -16,12 +16,32 @@ extern "C" {
 #include "SystemConfig.h"
 
 
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef enum
```
rickyelopez, line 23: we should eliminate these dependencies on the HAL. No need to use `GPIO_PIN_RESET` and `SET`, just use `0U` and `1U`. Don't just typedef HW_GPIO_TypeDef, redefine the type here so that we have a reference.

components/steering_wheel/HW/include/HW_gpio.h
```diff
@@ -16,12 +16,32 @@ extern "C" {
...
+} HW_GPIO_Input_S;
+
+
 /******************************************************************************
  *            P U B L I C  F U N C T I O N  P R O T O T Y P E S
  ******************************************************************************/
 
 void MX_GPIO_Init(void);
 
+HW_GPIO_PinState_E HW_GPIO_ReadPin(const HW_GPIO_Input_S *input);
```
rickyelopez, line 44: nit: remove `HW_` prefix, just GPIO is fine I think

components/steering_wheel/include/IO/IO_Digital.h
```diff
@@ -0,0 +1,48 @@
...
+ *      typedef'd IO_Digital_Flag_t must be modified to be congruent with the new system
+ */
+
+#pragma once
+
+/******************************************************************************
+ *                             I N C L U D E S
+ ******************************************************************************/
+
+#include "HW_gpio.h" /**< Needed for GPIO_TypeDef */
```
rickyelopez, line 14: remove comment, doesn't add value

rickyelopez:
This is looking good, really close I think

components/steering_wheel/include/IO/IO_Digital.h
JoshLafleur: Yes, I omitted them because they are handled in IO_Analog

components/steering_wheel/HW/HW_gpio.c
JoshLafleur: Should we not consider long term maintainability of these projects internally as well?

components/steering_wheel/HW/include/HW_gpio.h
JoshLafleur: I added HW_ to explicitly show that it is tied to the firmware interface. I think its beneficial long term just as HAL in regular stm32 programming shows its interfacing with hardware. Thoughts?

rickyelopez: I think GPIO already indicates that pretty clearly

components/steering_wheel/include/IO/IO_Digital.h
rickyelopez: gotcha. Those will probably be digital for now, I added analog functionality as a new feature but we'll start with digital. Anyway doesn't really matter we can fix it later

Issue #6, opened 2022-05-03
STW - Implement CAN Receiver


rickyelopez:
relates to #5 

The next step for the steering wheel is to implement the CAN receiver so that it can receive CAN messages from the other ECUs on the car - namely, the Motec.

Some thoughts
- Figure out how to set filters on the RX mailboxes
- Figure out if the peripheral can be configured to dump mailboxes into a FIFO
  - If not (which is most likely), I think we should add code to do that. Pretty sure that for the number of messages we'll be receiving and their speed, a FIFO should be more than sufficient.
- Figure out the signal access architecture
  - best way to abstract the receiver so that we can get it to call the receivers in a loop? Maybe make an unpack table the same way we have a pack table in the transmitter?
  - where to store the data?
  - typed access and raw access
  - atomic access to all the signals in the message?
- Add a `CANIO-rx` module which runs at each refresh rate and unpacks the messages for that rate
  - the rest of the modules that run afterwards should all have access to the most recent data

PR #10, opened 2022-06-21
CAN RX Implementation and Drawing improvements

rickyelopez:
Implements #6 

Implement CAN message reception. First step towards getting a full CAN receiver set up.
Also improves drawing and implements some of the pages that the old steering wheel had.


components/steering_wheel/include/Utility.h
```diff
@@ -55,3 +66,118 @@ static inline uint16_t u32CountLeadingZeroes(uint32_t x) { uint16_t c = 0U; whil
             (bit) = false;              \
         }                               \
     } while (zero())
+
+
+// FLAGS
```
rickyelopez, line 71: @JoshLafleur FYI this is the API I was referring to that you could use for the IO stuff

JoshLafleur: Regarding what we spoke about for the bitfield correct? This is the workaround to not being able to reference elements of a bitfield? To be frank I am a bit confued on the conversion from name to flag, the name would it be a define? or one of the elements of the bitfield? Would like to see more the workings of this

rickyelopez: Discussed offline

rickyelopez:
@JoshLafleur want to take a look? Big PR so no worries if not, we can just merge.

JoshLafleur (approved):

JoshLafleur:
@rickyelopez Want to update to current master and merge?

rickyelopez:
Confirmed builds still pass after rebasing. Not able to confirm that it actually works since I don't have any hardware with me right now. I think we should still merge, I can fix it later if it breaks anything

PR #12, opened 2022-06-23
Adding support to halt openocd or gdb on startup

JoshLafleur:
As spoken with @rickyelopez, this gives the option to append --halt to openocd or gdb to break upon startup

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -30,6 +30,8 @@ AddOption("--cleanbuild", dest="cleanbuild", action="store_true")
 AddOption("--pio_args", dest="pio_args", type="string", action="store", default="")
 # add option to build without debug info
 AddOption("--release", dest="release", action="store_true")
+# add option to halt openocd/gdb on startup
+AddOption("--halt", dest="halt", action="store_true")
 
 
 def pio_build_cmd(source, **_):
--- a/components/steering_wheel/SConscript
+++ b/components/steering_wheel/SConscript
@@ -241,12 +241,30 @@ env.Alias("cdb", env.CompilationDatabase())
 env.Alias("upload", env.flash(source="build/steering_wheel.bin"))
 env.Alias("openocd", env.openocd_srv(interface=OPENOCD_INTERFACE, board=OPENOCD_BOARD))
 
+# give support for halting on gdb/openocd start
+mon_reset_cmd = '-ex "monitor reset'
+if GetOption("halt"):
+    mon_reset_cmd += 'halt"'
+else:
+    mon_reset_cmd += '"'
+
+env.Alias(
+    "flash-debug",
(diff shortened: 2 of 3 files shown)


site_scons/site_tools/st-flash.py
```diff
@@ -14,6 +14,11 @@ def generate(env):
             "st-flash write $SOURCE 0x08000000"
         )
     )
+    env['BUILDERS']['flash_dbg'] = SCons.Builder.Builder(
+        action = SCons.Action.Action(
+            "st-flash --hot-plug write $SOURCE 0x08000000"
```
JoshLafleur, line 19: --hot-plug does not restart the MCU upon finishing FLASH-ing

components/steering_wheel/SConscript
```diff
@@ -241,19 +241,38 @@ env.Alias("cdb", env.CompilationDatabase())
...
 
+# give support for halting on gdb/openocd start
+mon_reset_cmd = '-ex "monitor reset'
+if GetOption("halt"):
+    mon_reset_cmd += 'halt"'
+else:
+    mon_reset_cmd += '"'
+
+env.Alias(
+    "flash-debug",
```
JoshLafleur, line 252: Flashes the MCU without restarting and then loads gdb with MCU halted

site_scons/site_tools/st-flash.py
rickyelopez: hmmmm weird, on my end the way it's implemented now already doesn't reset the MCU, I'm doing it manually as part of the gdb init script

JoshLafleur:
Odd, in that case on your end could you try a test? Try scons —target=stw flash with MX_gpio_init (iirc) in SystemManager.c main  commented out then again uncommented and let me know. For me it had always reset and started running the mcu. Either way do you think its a good option to have included long term for analysis of flash upon load?

SConstruct
```diff
@@ -30,6 +30,8 @@ AddOption("--cleanbuild", dest="cleanbuild", action="store_true")
 AddOption("--pio_args", dest="pio_args", type="string", action="store", default="")
 # add option to build without debug info
 AddOption("--release", dest="release", action="store_true")
+# add option to halt openocd/gdb on startup
+AddOption("--halt", dest="halt", action="store_true")
```
rickyelopez, line 34: Not sure I love this. You can add this param without one of the gdb targets and it will do nothing. I think it might be better to just add a new alias in the sconscript that halts the chip

components/steering_wheel/SConscript
rickyelopez: do we need this _and_ the `--halt` arg? Feels like just this is sufficient? Also I think it should be called `openocd-gdb-halt` since it does the same thing as `openocd-gdb` but also halts the chip

PR #14, opened 2022-07-10
SW-2 - Rewrite InfoDots and implement Pills

rickyelopez:


rickyelopez:
Merged into #10, will handle there.

PR #17, opened 2022-08-23
Steering Wheel Documentation WIP

JoshLafleur:


components/steering_wheel/doxygen_stw.conf
```diff
@@ -0,0 +1,2658 @@
...
+
+DOXYFILE_ENCODING      = UTF-8
+
+# The PROJECT_NAME tag is a single word (or a sequence of words surrounded by
+# double-quotes, unless you are using Doxywizard) that should identify the
+# project for which the documentation is generated. This name is used in the
+# title of most generated pages and in a few other places.
+# The default value is: My Project.
+
+PROJECT_NAME           = "My Project"
```
rickyelopez, line 35:
```suggestion
PROJECT_NAME           = "Steering Wheel"
```

rickyelopez (approved):
idk what other options should be changed in the doxygen config and tbh 2000 lines is a lot to go through lol

JoshLafleur:
> idk what other options should be changed in the doxygen config and tbh 2000 lines is a lot to go through lol

HAHAHA no I get it dont worry. I am using kind of template changes applied to the standard doxygen template but tailored to the project. Shouldn't have requested review on this PR as its still WIP sorry for that

components/steering_wheel/SConscript
```diff
@@ -229,6 +229,10 @@ objs.extend(compile_objects(env, project_source_files))
 
 ## done compiling
 
+doc = GetOption("doc")
+if doc:
+    Execute(f"doxygen doxygen_stw.conf")
```
rickyelopez, line 234: seems like you added a doxygen tool to `site_scons` in your other PR, why aren't you using it here?

Also, this should really be an alias rather than an execute
`Alias("docs", env.dox())` replace second parameter with the correct call into the tool you added

JoshLafleur: I actually wanted to talk about this with you. When I had that as the implementation (Why the dox.py exists), it would always execute, so I am a bit stuck with that there where it doesnt really make sense. Could you look into that and advise, as I am less familiar with SCons. This was the only method where it worked appropriately

PR #20, opened 2023-09-01
delete some useless test projects

rickyelopez:


rickyelopez:
Overriding merge requirements since this is just deleting some useless test projects

PR #21, opened 2023-09-01
STM32 Bootloader

rickyelopez:
Getting started on a bootloader for the STM32F103

So far, we're at around 7k of flash used for this. The original stm32duino bootloader (which used USB and DFU) was at about 8k.

To-do list:
- [x] get bootloader to boot
- [x] get bootloader to boot an app
- [x] get bootloader to talk on CAN 
- [x] bring up ISOTP
- [x] bring up UDS
- [ ] bring up UDS services
  - [x] tester present
  - [x] ecu reset
  - [x] routine control
  - [ ] download sequence
- [x] write flash manipulation code
- [x] write CRC calculation and comparison code
- [ ] Build a bootloader updater so that the bootloader can also be updated over CAN (this can technically actually happen separately

Not to mention, the other side of this which is to build out a full suite of tooling to actually leverage these capabilities on a computer. I have started writing a bit of code for this, since I needed it to test some of the above functions. There's a LOT to do here though.


components/bootloaders/STM/stm32f103/include/CAN.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
```
JoshLafleur, line 1: (nit) Needs file documentation and headers

components/bootloaders/STM/stm32f103/include/HW_CAN.h
```diff
@@ -0,0 +1,318 @@
...
+{
+    uint16_t                 id;
+
+    CAN_IdentifierLen_E      IDE;
+    CAN_RemoteTransmission_E RTR;
+
+    uint8_t                  lengthBytes;
+    CAN_data_T               data;
+    CAN_TxMailbox_E          mailbox;
+} CAN_TxMessage_t;
```
JoshLafleur, line 223: Keep consistent capitalization *_T

components/bootloaders/STM/stm32f103/include/HW.h
```diff
@@ -0,0 +1,265 @@
...
+    volatile uint16_t DR8;       // Data register 8
+    const uint16_t    RESERVED9;
+    volatile uint16_t DR9;       // Data register 9
+    const uint16_t    RESERVED10;
+    volatile uint16_t DR10;      // Data register 10
+    const uint16_t    RESERVED11;
+    volatile uint32_t RTCCR;     // RTC control register
+    volatile uint32_t CR;        // Control register
+    volatile uint32_t CSR;       // Control and status register
+} bkp_reg_map;
```
JoshLafleur, line 133: Consistent naming

components/bootloaders/STM/stm32f103/include/HW.h
```diff
@@ -0,0 +1,265 @@
...
+    USBWakeUp_IRQn        = 42,                   // USB Device WakeUp from suspend through EXTI Line Interrupt
+} IRQn_Type;
+
+
+// Power interface register map.
+typedef struct pwr_reg_map
+{
+    volatile uint32_t CR;    // Control register
+    volatile uint32_t CSR;   // Control and status register
+} pwr_reg_map;
```
JoshLafleur, line 231: Consistent naming

components/bootloaders/STM/stm32f103/include/HW_GPIO.h
```diff
@@ -0,0 +1,84 @@
...
+    GPIO_CFG_OUTPUT_OPEN_DRAIN = 0b01,
+} GPIO_reg_cnf;
+
+typedef enum
+{
+    GPIO_MODE_INPUT               = 0b00,
+    GPIO_MODE_OUTPUT_MEDIUM_SPEED = 0b01,
+    GPIO_MODE_OUTPUT_LOW_SPEED    = 0b10,
+    GPIO_MODE_OUTPUT_HIGH_SPEED   = 0b11,
+} GPIO_reg_mode;
```
JoshLafleur, line 38: Consistent naming

components/bootloaders/STM/stm32f103/include/HW_GPIO.h
```diff
@@ -0,0 +1,84 @@
...
+
+typedef enum
+{
+    GPIO_PULL_DOWN = 0U,
+    GPIO_PULL_UP,
+} GPIO_pull_dir;
```
JoshLafleur, line 44: Consistent naming

components/bootloaders/STM/stm32f103/include/HW_GPIO.h
```diff
@@ -0,0 +1,84 @@
...
+
+typedef struct
+{
+    uint32_t      port;
+    uint8_t       pin;
+    bool          alternate_function;
+    GPIO_reg_mode mode;
+    GPIO_reg_cnf  config;
+    GPIO_pull_dir pullDirection;
+} GPIO_config;
```
JoshLafleur, line 54: Consistent naming

components/bootloaders/STM/stm32f103/include/Utilities.h
```diff
@@ -0,0 +1,151 @@
+#pragma once
+#include "HW.h"
+#include "Types.h"
+
+#define UNUSED(x)                ((void)x)
+#define SATURATE(x, min, max)    (x > max) ? (max) : ((x < min) ? (min) : (x))
+// Count number of items in array
+#define COUNTOF(x)           ((uint16_t)(sizeof(x) / sizeof(x[0]))) // return size of an array as uint16
+
+// what is this?
```
JoshLafleur, line 10: Reading material:
https://lxr.linux.no/#linux+v2.6.31/include/linux/compiler-gcc.h#L12
https://en.wikipedia.org/wiki/Memory_ordering
Discussion: https://stackoverflow.com/questions/14950614/working-of-asm-volatile-memory

It seems it stops memory access ordering optimization at compile time. Helps make a stricter memory access flow, however does not prevent against CPU instruction optimization

components/bootloaders/STM/stm32f103/include/Utilities.h
JoshLafleur, line 1: Have the Atomic instructions been tested?

components/heartbeat/STM32F103C8.ld
```diff
@@ -0,0 +1,139 @@
...
+ENTRY(Reset_Handler)
+
+/* Highest address of the user mode stack */
+_estack = ORIGIN(RAM) + LENGTH(RAM);    /* bottom of stack (grows down) */
+
+__STACK_SIZE = 0x400;
+__HEAP_SIZE = 0x200;
+
+__RAM_SIZE = 20K;
+__FLASH_SIZE = 58K;
```
JoshLafleur, line 16: I think we would benefit from allocation a 1kB nvm area at the back of flash. Will be useful to integrate it from the start for all of the bms and other systems

JoshLafleur (commented):
Files needs headers and documentation, but I definetly like the idea of going towards fully custom firmware custom, really great stuff going and getting all of that laid out!

I like the starting point and direction, really awesome of you to take that lead

rickyelopez:
> Files needs headers and documentation, but I definetly like the idea of going towards fully custom firmware custom, really great stuff going and getting all of that laid out!
> 
> I like the starting point and direction, really awesome of you to take that lead

lots of cleanup to be done before this is anywhere near merge-able (not the least of which is documentation), namely getting it to actually do bootloader things lol

components/bootloaders/STM/stm32f103/include/CAN.h
rickyelopez: yeah all these files are going to need that

components/bootloaders/STM/stm32f103/include/HW_CAN.h
rickyelopez: I think we should stick to lowercase t, like with all the standard types (i.e `uint8_t`)

components/bootloaders/STM/stm32f103/include/HW.h
rickyelopez: yeah this stuff is all copy-pasted, I'll rename everything before it's ready to merge. More important to get it working right now though

components/bootloaders/STM/stm32f103/include/Utilities.h
rickyelopez: nice, very interesting

components/bootloaders/STM/stm32f103/include/Utilities.h
rickyelopez: They come straight from the cortex-m3 library files, they're intrinsics. We'll have to test them to confirm, though

components/heartbeat/STM32F103C8.ld
rickyelopez: once the bootloader size is stabilized we can allocate an extra page at the beginning of the flash for nvm. iirc the page sizes are bigger at the end of flash than at the beginning, so ideally we want to allocate them at the beginning of flash for the bootloader so they don't consume too much application space. The apps can use whatever is left at the end of flash for nvm, unless there's less than a page. If we run into that, we could also explore having the app's nvm pages at the front too

rickyelopez:
Really cool progress!

image

CAN messages being sent out from four different periodic functions which are driven by the timer. Also, UDS frame being sent out (just a simple Tester Present ACK) by the library I started writing, and you can see it is decoded correctly as a UDS frame!

rickyelopez:
This PR is starting to get a little big, and it will only get worse. I propose we merge this as-is since most (if not all) of the behavior is confirmed working, and then keep working on it in a subsequent PR.

rickyelopez:
More good progress

image

Got a full UDS transaction to work!
`Send from PC -> receive in the bootloader -> respond -> PC recognizes response`

Cool!

rickyelopez:
With this latest push, app CRC generation (at build time) and verification (in the bootloader) are working!!

image

In the above photo, you can see the periodics spitting out data, but also on address 0x402 you can see the CRC as read from the app's flash region, as well as the app CRC as calculated by the bootloader. As you can see, they match! Very cool!

I think the only big part of this that is left to do is the actual app download, the verification of which will be hugely facilitated by having a functional CRC comparison!

components/heartbeat/STM32F103C8.ld
rickyelopez: I was wrong about this. On this chip, all flash pages are 1k. We can allocate wherever, but probably not worth doing until we actually have a need for nvm

rickyelopez:
Ok, this is ready to merge. Almost everything we need, with the notable exception of app downloads, is working. Let's merge this so we can progress with other work. I'll do app downloads in a separate pr

JoshLafleur (approved):

PR #22, opened 2023-09-02
Merge Safety Board program into Master

JoshLafleur:
Working implementation of CFR23 safety board for BMS


components/safety_board/HW/HW_can.c
```diff
@@ -0,0 +1,467 @@
+/**
+ * @file HW_can.c
+ * @brief  Source code for the CAN hardware
+ * @author Joshua Lafleur (josh.lafleur@outlook.com)
+ * @version 0.1
+ * @date 2023-01-15
```
rickyelopez, line 6: don't put versions or dates anywhere. These files live in a version controlled repo, these comments don't add value but do add confusion (when people forget to update them)

components/safety_board/HW/HW_can.c
```diff
@@ -0,0 +1,467 @@
...
+ */
+
+/******************************************************************************
+ *                             I N C L U D E S
+ ******************************************************************************/
+
+#include "HW_can.h"
+
+#include "CAN/CAN_types.h"
+#include "include/ErrorHandler.h"
```
rickyelopez, line 16: no need to add `include/` here

components/safety_board/HW/HW_can.c
```diff
@@ -0,0 +1,467 @@
...
+#include "SystemConfig.h"
+#include "string.h"
```
rickyelopez, line 18: this is a system header, include should be `#include <string.h>`

components/safety_board/HW/HW_can.c
```diff
@@ -0,0 +1,467 @@
...
+static void CAN_RxMsgPending_ISR(CAN_HandleTypeDef* canHandle, CAN_RxFifo_E fifoId)
+{
+    uint8_t data[8];
+    CAN_RxHeaderTypeDef header;
+
+    if (canHandle == &hcan)
+    {
+        HAL_CAN_GetRxMessage(canHandle, fifoId, &header, &data[0]);
+
+        if (header.StdId == BMS_STATUS_ID)
```
rickyelopez, line 307: this is high-level code, it does not belong in the low-level file.

components/safety_board/HW/HW_gpio.c
```diff
@@ -0,0 +1,139 @@
...
+}
+
+/**
+ * @brief  Interrupt to cycle state
+ *
+ * @param GPIO_Pin
+ */
+void HAL_GPIO_EXTI_Callback(uint16_t GPIO_Pin)
+{
+    if (GPIO_Pin == TSMS_CHG_Pin)
```
rickyelopez, line 95: high-level code in low-level file

components/safety_board/HW/HW_gpio.c
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+    else if (GPIO_Pin == GPIO_PIN_0)
+    {
+        // Only set through internal software interrupts
+        SYS_SAFETY_CycleState();
+    }
+}
+
+uint8_t HW_GPIO_ValidateInputs(void)
```
rickyelopez, line 124: same here

components/safety_board/HW/HW_msp.c
```diff
@@ -0,0 +1,34 @@
...
+ ******************************************************************************/
+
+/**
+ * @brief  Called after HAL Init
+ */
+void HAL_MspInit(void)
+{
+    __HAL_RCC_AFIO_CLK_ENABLE();
+    __HAL_RCC_PWR_CLK_ENABLE();
+    HAL_NVIC_SetPriority(PendSV_IRQn, 15, 0);
```
rickyelopez, line 29: this should probably be higher priority? I believe a higher number here means lower priority

components/safety_board/HW/include/HW_gpio.h
```diff
@@ -0,0 +1,25 @@
...
+#include "SystemConfig.h"
+
+
+/******************************************************************************
+ *            P U B L I C  F U N C T I O N  P R O T O T Y P E S
+ ******************************************************************************/
+
+void HW_GPIO_Init(void);
+void HW_CONTACTORS_CloseAll(void);
+uint8_t HW_GPIO_ValidateInputs(void);
```
rickyelopez, line 24: high-level code in low-level file

components/safety_board/HW/include/SystemConfig.h
```diff
@@ -0,0 +1,86 @@
...
+ ******************************************************************************/
+
+#define BMS_STATUS_ID 0x037
+#define BMS_FLAG_BYTE 0x00
+#define BMS_ISREADY_BIT  0x02
+#define BMS_ISCHARGE_BIT 0x03
+#define BMS_ISERROR_BIT 0x07
+#define BMS_VOLTAGE_ID 0x32
+#define BMS_BATT_VOLTAGE_MSB 0x01
+#define BMS_BATT_VOLTAGE_LSB 0x00
```
rickyelopez, line 31: why do you need these?

components/safety_board/HW/include/SystemConfig.h
```diff
@@ -0,0 +1,86 @@
...
+#include "stm32f1xx.h"
+
+#include "ErrorHandler.h"
+
+
+/******************************************************************************
+ *                              D E F I N E S
+ ******************************************************************************/
+
+#define BMS_STATUS_ID 0x037
```
rickyelopez, line 24: these being hard-coded is fine for now, but this should come from the network generator tool eventually, and you should add the definition for these messages to it in this PR as well (as a source of truth that is human readable)

components/safety_board/include/SYS_Vehicle.h
```diff
@@ -0,0 +1,74 @@
...
+ *                              D E F I N E S
+ ******************************************************************************/
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef enum
+{
+  IMD_STATUS = 0x00,
```
rickyelopez, line 28: enums should follow the same pattern and only diverge at the end. 

eg, this should be
```c
STATUS_INDEX_IMD = 0x00,
STATUS_INDEX_BMS,
```
etc.

Same comment on the other enums below

components/safety_board/include/SYS_Vehicle.h
```diff
@@ -0,0 +1,74 @@
...
+  BMS_STATUS,
+  BMS_DISCHARGE_STATUS,
+  BMS_CHARGE_STATUS,
+  BMS_ERROR_STATUS,
+  TSMS_STATUS,
+  TS_STATUS,
+  STATUS_COUNT,
+} STATUS_INDEX_E;
```
rickyelopez, line 36: defs not all caps for enum types
maybe `statusIndex_E`?
Same below

components/safety_board/src/SYS_Vehicle.c
```diff
@@ -0,0 +1,255 @@
...
+ ******************************************************************************/
+
+static uint8_t          status_flags      = 0;
+static VEHICLE_STATES_E vehicle_state     = VEHICLE_INIT;
+static int16_t          ts_voltage        = 0;    // Stored at 0.1 V increments
+static int16_t          batt_voltage      = 0;    // Stored at 0.1 V increments
+static uint32_t         last_ts_message   = 0;
+static uint32_t         last_batt_message = 0;
+static uint8_t          imd_duty_cycle = 0;
+static uint8_t          imd_freq = 0;
```
rickyelopez, line 33: these should all be in a module-level struct

components/safety_board/src/SYS_Vehicle.c
```diff
@@ -0,0 +1,255 @@
+/**
```
rickyelopez, line 1: Why `SYS_Vehicle.c/h`? Why SYS? What does this file actually containt?

components/safety_board/src/SYS_Vehicle.c
```diff
@@ -0,0 +1,255 @@
...
+ *          P R I V A T E  F U N C T I O N  P R O T O T Y P E S
+ ******************************************************************************/
+
+bool SYS_Validate(void);
+
+/******************************************************************************
+ *                       P U B L I C  F U N C T I O N S
+ ******************************************************************************/
+
+void SYS_SAFETY_Init(void)
```
rickyelopez, line 46: way too much all-caps in this naming convention. If this is safety related, maybe just
`SFTY_init`?
A common and (in my opinion) nice standard is all-caps module specifier (`SFTY` in this case) and then camel-case after that.

components/safety_board/src/SYS_Vehicle.c
```diff
@@ -0,0 +1,255 @@
...
+
+void SYS_CONTACTORS_OpenAll(void)
+{
+    HAL_GPIO_WritePin(AIR_Port, AIR_Pin, GPIO_PIN_RESET);
+    HAL_GPIO_WritePin(PCHG_Port, PCHG_Pin, GPIO_PIN_RESET);
+}
+
+bool SYS_SAFETY_GetStatus(STATUS_INDEX_E status)
+{
+    return (status_flags & (0x01 << status));
```
rickyelopez, line 77: why not use the Flags module I build in the steering wheel code? We could move that out to the libs dir

components/safety_board/src/SYS_Vehicle.c
```diff
@@ -0,0 +1,255 @@
...
+    data.u8[6] = imd_duty_cycle;
+    data.u8[7] = imd_freq;
+    
+    extern CAN_HandleTypeDef hcan;
+    static uint8_t rst_counter = 0;
+
+    if (!CAN_sendMsgBus0(CAN_TX_PRIO_100HZ, data, SAFETY_BOARD_STATUS_ID, 8)
+        && rst_counter++ == 100)
+    {
+        HAL_NVIC_SystemReset();
```
rickyelopez, line 213: you're doing a whole board reset if CAN can't transmit? That seems wrong to me

components/safety_board/src/SYS_Vehicle.c
```diff
@@ -0,0 +1,255 @@
...
+        rst_counter = 0;
+    }
+
+}
+
+/******************************************************************************
+ *                     P R I V A T E  F U N C T I O N S
+ ******************************************************************************/
+
+bool SYS_Validate(void)
```
rickyelopez, line 230: I think all this logic needs to be broken up into at least one state machine. I can help define it

site_scons/components.yaml
```diff
@@ -2,3 +2,8 @@ stw:
   path: components/steering_wheel/
 network:
   path: network/
+
+sfb:
+  path: components/safety_board/
+network:
```
rickyelopez, line 8: duplicated

rickyelopez:
Also, this definitely needs to be using FreeRTOS

PR #23, opened 2023-09-02
fix openocd

rickyelopez:


JoshLafleur (approved):
Looks good to me, not the most familiar with TCL but seems good

rickyelopez:
yeah this is just copy-pasted from their repo for the newer version of openOCD, nothing interesting here.

PR #24, opened 2023-11-06
Added a clean argument to buildroot.sh and modified the readme

JoshLafleur:
This was done to facilitate updating the container without having to search for the commands/workflows.

### DIFF
--- a/buildroot.sh
+++ b/buildroot.sh
@@ -1,5 +1,29 @@
 #!/bin/bash
+is_arg=false
 
-docker-compose pull
-docker-compose run --rm buildroot
+for args in "$@"
+do
+  is_arg=true
+  
+  case $args in
+    -h|--help)
+    echo "Buildroot help..."
+    echo "  --help          Displays this help message"
 
+  ;;
+  *)
+    echo "Unknown option: $args"; exit 1
+  ;;
+  ## If option with variable
+  #-p=*|--prefix=*)
+  #OPTION="${i#*=}"
+
+  #;;
+  esac
+done
+
+if [ "$is_arg" = false ]; then
+  docker-compose pull
+  docker-compose up -d
+  docker-compose run --rm buildroot
+fi
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
 # Firmware Repo
 
+## buildroot.sh commands
+- Running `./buildroot.sh` from the root of the repository will start the docker environment.
+  - `--help` can be appended to display buildroot.sh script options
 ## Will contain all of the firmware code


buildroot.sh
```diff
@@ -1,5 +1,29 @@
 #!/bin/bash
+is_arg=false
 
-docker-compose pull
-docker-compose run --rm buildroot
+for args in "$@"
+do
+  is_arg=true
 
+  if [ "$args" = "--help" ]; then
```
rickyelopez, line 8: this is not the right way to do this. See example here: https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash
using a switch statement

buildroot.sh
```diff
@@ -1,5 +1,29 @@
...
+  elif [ "$args" = "--clean" ]; then
+    echo "Cleaning system..."
+    docker-compose stop
+    docker-compose rm -f
+    docker-compose pull
+    docker-compose up -d
+  elif [ "$args" = "--setup" ]; then
+    echo "Setting up system..."
+    echo "This assumes that you already have docker, docker-compose, and git already installed"
+    git submodule update --init --recursive
```
rickyelopez, line 22: I don't think this is necessary. Why is this easier than just running the command directly?

buildroot.sh
```diff
@@ -1,5 +1,29 @@
...
+    echo "Buildroot help..."
+    echo "  --help          Displays this help message"
+    echo "  --clean         Stops the buildroot container, removes the container and its data, pulls the most recent version, and updates the container with the most recent image"
+    echo "  --setup         Setup environment dependancies"
+  elif [ "$args" = "--clean" ]; then
+    echo "Cleaning system..."
+    docker-compose stop
+    docker-compose rm -f
+    docker-compose pull
+    docker-compose up -d
```
rickyelopez, line 18: `stop` and `rm` shouldn't be necessary if the container was started with `--rm` (which, looking at this file, it is)

JoshLafleur:
Made changes based on comments and squashed to single commit for cleanliness of merge into master

buildroot.sh
```diff
@@ -1,5 +1,29 @@
...
+
+  #;;
+  esac
+done
+
+if [ "$is_arg" = false ]; then
+  docker-compose pull
+  docker-compose up -d
+  docker-compose run --rm buildroot
+fi
```
rickyelopez, line 29:
```suggestion
POSITIONAL_ARGS=()

print_help () {
    echo "Buildroot help..."
    echo "  --help          Displays this help message"
}

while [[ $# -gt 0 ]]; do
  case $1 in
    -h|--help)
        print_help()
        exit 0
        ;;
    # ...
    # can add other args here in the future
    # ...
    -*|--*) # catch any unknown args and exit
        echo "Unknown option $1"
        print_help()
        exit 1
        ;;
    *)
        POSITIONAL_ARGS+=("$1") # save positional arg
        shift # past argument
        ;;
  esac
done

set -- "${POSITIONAL_ARGS[@]}" # restore positional parameters

# one day we can do things here with positional args or params

# for now, just always start the container
docker-compose pull
docker-compose up -d
docker-compose run --rm buildroot
```

PR #28, opened 2023-12-17
Add BMS Worker to master

JoshLafleur:


components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -0,0 +1,95 @@
...
+    simpleFilter_S cell_voltages[ADC_CHANNEL_COUNT];
+    simpleFilter_S pack_voltage;
+} BMS_Data_S;
+
+typedef struct {
+    BMS_State_E state;
+    uint16_t balancing_cells;
+    uint16_t cell_faults;
+    uint16_t cell_voltages[CELL_COUNT];
+    uint16_t pack_voltage; /**< Stored in 1mV units */
```
rickyelopez, line 68: should use something like this for units:
```suggestion
    uint16_t pack_voltage; // [mV] pack voltage, precision of 1mV
```

components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -0,0 +1,95 @@
...
+    uint16_t calculated_pack_voltage; /**< Stored in 1mV units */
+    uint16_t max_voltage; /**< Stored in 0.0001 units */
+    uint16_t min_voltage; /**< Stored in 0.0001 units */
+    uint16_t avg_voltage; /**< Stored in 0.0001 units */
```
rickyelopez, line 72: use structs for clarity and to group relevant quantities:
```suggestion
    struct
    {
        uint16_t max; // [V] max voltage, precision of 0.0001 V
        uint16_t min; // [V] max voltage, precision of 0.0001 V
        uint16_t avg; // [V] max voltage, precision of 0.0001 V
    } voltage;
```

also, why not store these in floats?

components/bms_worker/include/Environment.h
```diff
@@ -0,0 +1,84 @@
...
+        int16_t ambient_temp; /**< Stored in 0.1 deg C */
+        uint16_t rh; /**< Stored in 0.01% RH */
+    } board;
+    struct {
+        int16_t cell_temps[CHANNEL_COUNT]; /**< Stored in 0.1 deg C */ 
+        int16_t max_temp;
+        int16_t min_temp;
+        int16_t avg_temp;
+    } cells;
+} Env_Variables_S;
```
rickyelopez, line 51: why? just put these in the module struct directly

components/bms_worker/include/FeatureDefs.h
```diff
@@ -0,0 +1,18 @@
...
+ ******************************************************************************/
+
+#include <stdbool.h>
+
+
+/******************************************************************************
+ *                              D E F I N E S
+ ******************************************************************************/
+
+#define FTR_HALL_EFFECT_PADDLES    true
```
rickyelopez, line 18: delete

components/bms_worker/include/Module.h
```diff
@@ -0,0 +1,66 @@
...
+#include "stdint.h"
+
+
+/******************************************************************************
+ *                              E X T E R N S
+ ******************************************************************************/
+
+/**< Modules */
+extern const ModuleDesc_S BMS_desc;
+extern const ModuleDesc_S Cooling_desc;
```
rickyelopez, line 25: module names should be either acronyms or abbreviations. eg. COOL and ENV instead of Cooling and Environment. This will make it easier to name variables and functions in such a way that the namespace is clear

components/bms_worker/include/Sys.h
```diff
@@ -0,0 +1,79 @@
...
+typedef uint8_t BMS_Errors_t; 
+
+typedef enum {
+    BMS_ERROR = 0x00,
+    CELL_TEMP_ERROR,
+    RH_ERROR,
+    AMBIENT_TEMP_ERROR,
+    BOARD_TEMP_ERROR,
+    ERRORS_CNT,
+} Sys_Errors_E;
```
rickyelopez, line 38: I've noticed this everywhere, but all the naming should be using lowerCamelCase, where snake_case should be used to separate parts of the name. For example, this should be SYS_errors. If the name were longer, you'd use camel case. Eg. SYS_errorsPlusMoreWords, etc.

components/bms_worker/include/Sys.h
```diff
@@ -0,0 +1,79 @@
...
+ ******************************************************************************/
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef uint8_t BMS_Errors_t; 
+
+typedef enum {
+    BMS_ERROR = 0x00,
```
rickyelopez, line 32: the part of the enum that is different should be at the end, so the majority of each one is consistent. Also, namespacing. Eg. in this case you could do SYS_ERROR_BMS, SYS_ERROR_CELL_TEMP, etc.

components/bms_worker/include/Utility.h
```diff
@@ -0,0 +1,251 @@
...
+#include <stdint.h>
+
+/******************************************************************************
+ *                               M A C R O S
+ ******************************************************************************/
+
+// Count number of items in array
+#define COUNTOF(x)    ((uint16_t)(sizeof(x) / sizeof(x[0])))    // return size of an array as uint16
+// Count leading zeroes
+static inline uint16_t u32CountLeadingZeroes(uint32_t x)
```
rickyelopez, line 20: there's an intrinsic for this, I'll dig it up at some point.

components/bms_worker/src/Cooling.c
```diff
@@ -0,0 +1,156 @@
...
+/**
+ * @brief  Cooling Module Init function
+ */
+static void Cooling_Init()
+{
+    COOLING.state      = COOLING_INIT;
+    COOLING.percentage = 0;
+    FANS_Init();
+
+    COOLING.state = COOLING_INIT;
```
rickyelopez, line 75: duplicate of line 71

components/bms_worker/src/Cooling.c
```diff
@@ -0,0 +1,156 @@
...
+        case COOLING_INIT:
+            step += 20;
+            COOLING.percentage = (step <= 100) ? step : 200 - step;
+
+            if (step >= 200)
+                COOLING.state = COOLING_OFF;
+            break;
+        case COOLING_ON:
+        case COOLING_OFF:
+            if (ENV.values.cells.max_temp < 350)
```
rickyelopez, line 96: should do a simple linear interpolation instead of this. Or maybe a define a mathematical function that translates temp to percentage.

components/bms_worker/src/Cooling.c
```diff
@@ -0,0 +1,156 @@
...
+            }
+            else if (ENV.values.cells.max_temp > 400)
+            {
+                COOLING.state = COOLING_ON;
+                COOLING.percentage = 40;
+            }
+            else if (ENV.values.cells.max_temp >= 350)
+            {
+                COOLING.state = COOLING_ON;
+                COOLING.percentage = 20;
```
rickyelopez, line 123: what do these states actually accomplish? seems like cooling state just represents whether percentage is 0 or not?

components/bms_worker/src/Environment.c
```diff
@@ -0,0 +1,174 @@
...
+                if (ltc_chip.temps[i] < tmp_min) tmp_min = ltc_chip.temps[i];
+                if (ltc_chip.temps[i] > tmp_max) tmp_max = ltc_chip.temps[i];
+                tmp_avg += ltc_chip.temps[i];
+            }
+            
+            tmp_avg /= CHANNEL_COUNT;
+
+            ENV.values.cells.avg_temp = (uint16_t)tmp_avg;
+            ENV.values.cells.max_temp = tmp_max;
+            ENV.values.cells.min_temp = tmp_min;
```
rickyelopez, line 122: I think it would be a lot more clear if these quantities were floats. Also easier to work with elsewhere

site_scons/components.yaml
```diff
@@ -2,7 +2,9 @@ bl:
   path: components/bootloaders/
 heartbeat:
   path: components/heartbeat/
-network:
-  path: network/
 stw:
   path: components/steering_wheel/
+bmsw:
```
rickyelopez, line 7: pls alphabetize

rickyelopez (commented):
overall looks great. Lots of housekeeping to do but there's a great foundation here

components/bms_worker/HW/HW_Fans.c
```diff
@@ -0,0 +1,147 @@
...
+    bool init_state = true;
+
+    current_state = OFF;
+    FANS_SetPower(0);
+    
+    if (init_state != true) 
+    {
+        Error_Handler();
+        return false;
+    }
```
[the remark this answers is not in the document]
JoshLafleur: rabbit hole its resolved

components/bms_worker/src/Cooling.c
JoshLafleur: Added states because in full blast theres a hysterisis to come out of it. Added ON and OFF more as information and forward compatibility if ever the cooling medium and control changes. For Cooling off can also integrate long term into low power modes if ever integrate power drawn from segments constantly

components/bms_worker/src/Environment.c
JoshLafleur: I try to keep it in stdint. I have no literature but as i understand soft floats are computationally heavy

components/bms_worker/HW/HW_Fans.c
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        HW_TIM1_setDuty((percentage < 50) ? 50 : percentage);
+        return;
+    }
+
+    /**< Handle non-linearity of optocoupler output for duty-cycle -> response */
+    if (percentage > 100)
+        percentage = 100;
+    else if (percentage <= 10 && percentage > 0)
```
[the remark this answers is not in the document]
JoshLafleur: This was on a1 design for non-linearity. with new digital isolator much better rise and fall times

components/bms_worker/HW/HW_Fans.c
```diff
@@ -0,0 +1,147 @@
...
+        {   current_state = STARTING;
+            start_time = HAL_GetTick();
+        }
+        else if (start_time + 500 < HAL_GetTick())
+        {
+            current_state = RUNNING;
+            start_time = 0;
+        }
+
+        HW_TIM1_setDuty((percentage < 50) ? 50 : percentage);
```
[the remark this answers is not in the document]
JoshLafleur: When starting up fans (and specifie in the datasheet), when starting a fan they sometimes need a higher command level to start it spinning. This was done for a set time on va1 however on va3 and the speed feedback now the command level drops when the feedback speed is above 250rpm.

components/bms_worker/HW/HW_Fans.c
```diff
@@ -0,0 +1,147 @@
...
+        return;
+    }
+
+    /**< Handle non-linearity of optocoupler output for duty-cycle -> response */
+    if (percentage > 100)
+        percentage = 100;
+    else if (percentage <= 10 && percentage > 0)
+        percentage = 10;
+    else if (percentage < 25 && percentage > 10)
+        percentage = 10 + (uint16_t) percentage * 5/25;
```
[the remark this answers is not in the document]
JoshLafleur: The optocoupler has a nonlinear response due to the relatively high rise and fall times. to get a linear fan control I made a transfer function to get better linear control, however on the newly designed circuit the hardware handles it and was removed.

components/bms_worker/HW/HW_Fans.c
```diff
@@ -0,0 +1,147 @@
+/**
```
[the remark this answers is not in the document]
JoshLafleur: As I understand this is a hardware driver and therefore is equivalent to firmware in a sense. You would recommend keeping it in the HW folder but just named Fans.c?

components/bms_worker/HW/HW_HS4011.c
```diff
@@ -0,0 +1,127 @@
+/**
+ * @file HW_HS4011.c
+ * @brief  Source file for HS4011 Relative Humidty/Temperature Sensor
+ * @author Joshua Lafleur (josh.lafleur@outlook.com)
+ * @version
+ * @date 2023-12-19
+ */
+
+#if defined(BMSW_BOARD_VA1)
```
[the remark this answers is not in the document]
JoshLafleur: The a1 version of the board. the va1 and va3 are different for the signals, pins, etc

components/bms_worker/HW/HW_HS4011.c
```diff
@@ -0,0 +1,127 @@
...
+extern HW_I2C_Handle_T i2c2;
+
+
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+/******************************************************************************
+ *                               M A C R O S
+ ******************************************************************************/
```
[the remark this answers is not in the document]
JoshLafleur: Yea i just copied them all when making the files helped at the start but turned into a pain. Will be revising

components/bms_worker/HW/HW_HS4011.c
```diff
@@ -0,0 +1,127 @@
...
+}
+
+bool HS4011_GetData()
+{
+    uint8_t rdata[5] = { 0 };
+
+    if (!HW_I2C_Master_Read(hs_chip.dev, (uint8_t*)&rdata, 5, 100))
+        return false;
+
+    hs_chip.data.measuring = false;
```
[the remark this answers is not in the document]
JoshLafleur: This handles a bus NACK

components/bms_worker/HW/HW_HS4011.c
```diff
@@ -0,0 +1,127 @@
...
+
+    reverse_bytes((uint8_t*)&rdata[0], 2);
+    reverse_bytes((uint8_t*)&rdata[2], 2);
+
+    hs_chip.data.rh   = (uint16_t)(((((uint32_t)rdata[1] << 8) | ((uint32_t)rdata[0])) * 10000) / 16383);
```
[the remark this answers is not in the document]
JoshLafleur: I detailed the information there, the casts are to avoid overflowing a 16bit

components/bms_worker/HW/include/HW_LTC2983.h
```diff
@@ -0,0 +1,104 @@
...
+
+} LTC_Config_S;
+
+typedef struct {
+    bool hard_fault;
+    bool soft_above;
+    bool soft_below;
+    bool valid;
+    int32_t result;
+    uint32_t raw;
```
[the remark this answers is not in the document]
JoshLafleur: It is for debugging. have not built out additional macros for added debuggin functionality

JoshLafleur:
I have updated the new version to this branch to not make changes to older code. I have merged it into bms_worker and will continue the review from here

components/bms_worker/HW/HW.c
```diff
@@ -0,0 +1,95 @@
...
+ * @brief  Source code for generic firmware functions
+ * @author Joshua Lafleur (josh.lafleur@outlook.com)
+ * @date 2023-12-28
+ */
+
+/******************************************************************************
+ *                             I N C L U D E S
+ ******************************************************************************/
+
+/**< Header file */
```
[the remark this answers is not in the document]
JoshLafleur: Still needs work but got a bulk of style changes done

components/bms_worker/include/BatteryMonitoring.h
JoshLafleur: once converted from the IO module it is kept in the highest level of reasonable precision (see diff between pack and individual cell)

components/bms_worker/HW/HW_Fans.c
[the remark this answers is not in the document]
rickyelopez: 1. all of this is firmware
2. the only thing in this file that interfaces with hardware is the command to set the pwm duty cycle
  - the pwm part of this should be moved to a hw_pwm.c file or similar, and this file should be moved out of the hw dir

PR #29, opened 2024-01-04
Peripheral DeInit Drivers

Gray-man:
Wrote DeInit drivers for HW.c, HW_gpio.c, HW_tim.c, HW_i2c.c, HW_can.c, HW_dma.c, HW_adc.c, and HW_spi.c


JoshLafleur:
Havent checked through all the HW files, but im pretty sure theres more todo. every file in theory in HW should have a change. Also need to deinitialize rcc and clocks

JoshLafleur:
Add Doxygen comments to each function you wrote, detailing reutrn values (they should all return a bool), and any parameters

PR #36, opened 2024-03-14
pull request to User/lt/bms_worker_cells from user/lt/bms_worker_2

tonynokart:
for calculations of current limits for charge/discharge due to temperature

### DIFF
--- a/components/bms_worker/include/BatteryMonitoring.h
+++ b/components/bms_worker/include/BatteryMonitoring.h
@@ -65,6 +65,9 @@ typedef struct
     float       chargeLimit;
     float       dischargeLimit;
 
+    float       heatDischargeCurrentLimit;
+    float       heatChargeCurrentLimit;
+
     struct
     {
         uint16_t max;
@@ -104,3 +107,5 @@ void BMS_measurementComplete(void);
 void BMS_ChargeLimit(uint16_t relativeSoC);
 void BMS_DischargeLimit(uint16_t relativeSoC);
 
+void BMS_HeatCurrentChargeLimit(int16_t cellTemp);
+void BMS_HeatCurrentDischargeLimit(int16_t cellTemp);
--- a/components/bms_worker/include/BatteryMonitoring.h
+++ b/components/bms_worker/include/BatteryMonitoring.h
@@ -103,3 +103,4 @@ void BMS_setOutputCell(MAX_selectedCell_E cell);
 void BMS_measurementComplete(void);
 void BMS_ChargeLimit(uint16_t relativeSoC);
 void BMS_DischargeLimit(uint16_t relativeSoC);
+
--- a/components/bms_worker/include/Environment.h
+++ b/components/bms_worker/include/Environment.h
@@ -75,6 +75,8 @@ typedef struct
 typedef struct
 {
     Environment_State_E state;
+    float       heatDischargeCurrentLimit;
+    float       heatChargeCurrentLimit;
     struct
     {
         struct
@@ -97,3 +99,10 @@ typedef struct
  ******************************************************************************/
 
 extern ENV_S ENV;
+
+/******************************************************************************
+ *            P U B L I C  F U N C T I O N  P R O T O T Y P E S
+ ******************************************************************************/
+
+void ENV_HeatCurrentChargeLimit(int16_t cellTemp);
+void ENV_HeatCurrentDischargeLimit(int16_t cellTemp);
--- a/components/bms_worker/include/Environment.h
+++ b/components/bms_worker/include/Environment.h
@@ -75,8 +75,6 @@ typedef struct
 typedef struct
 {
     Environment_State_E state;
-    float       heatDischargeCurrentLimit;
-    float       heatChargeCurrentLimit;
     struct
     {
         struct
@@ -100,9 +98,3 @@ typedef struct
 
 extern ENV_S ENV;
 
-/******************************************************************************
- *            P U B L I C  F U N C T I O N  P R O T O T Y P E S
- ******************************************************************************/
-
-void ENV_HeatCurrentChargeLimit(int16_t cellTemp);
-void ENV_HeatCurrentDischargeLimit(int16_t cellTemp);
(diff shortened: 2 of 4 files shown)


components/bms_worker/src/Environment.c
JoshLafleur, line 1: Remove changes to the file. Less diffs = more better

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -417,3 +420,25 @@ void BMS_DischargeLimit(uint16_t relativeSoC) {
         BMS.dischargeLimit = 2.25f*relativeSoC; //linear function for the last 20% of discharge
     }
 }
+
+void BMS_HeatCurrentDischargeLimit(int16_t cellTemp) {
```
JoshLafleur, line 424: SHould be lowerCamelCase

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -417,3 +420,25 @@ void BMS_DischargeLimit(uint16_t relativeSoC) {
...
+    if (cellTemp >= 48) {
+        BMS.heatChargeCurrentLimit    = -3.75f * cellTemp + 225;
+    } else if (cellTemp > 60) {
+        BMS.heatDischargeCurrentLimit = 0;
+    } else {
+        BMS.heatDischargeCurrentLimit = MAX_CONTINOUS_DISCHARGE_CURRENT;
+    }
+}
+
+void BMS_HeatCurrentChargeLimit(int16_t cellTemp) {
```
JoshLafleur, line 434: lowerCamelCase gain. IE BMS_heatCurrentChargeLimit

components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -103,3 +106,6 @@ void BMS_setOutputCell(MAX_selectedCell_E cell);
 void BMS_measurementComplete(void);
 void BMS_ChargeLimit(uint16_t relativeSoC);
 void BMS_DischargeLimit(uint16_t relativeSoC);
+
+void BMS_HeatCurrentChargeLimit(int16_t cellTemp);
```
JoshLafleur, line 110: lowerCamelCase

components/bms_worker/src/BatteryMonitoring.c
JoshLafleur: If you pass the cellTemp as an arg, you should return the calculated value. if not an argm, then you should take and put to the struct from the function

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -417,3 +420,25 @@ void BMS_DischargeLimit(uint16_t relativeSoC) {
...
+    if (cellTemp >= 48) {
```
JoshLafleur, line 435: logic is faulty, if temp is above 60 you will never reach line 438

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -417,3 +420,25 @@ void BMS_DischargeLimit(uint16_t relativeSoC) {
         BMS.dischargeLimit = 2.25f*relativeSoC; //linear function for the last 20% of discharge
     }
 }
+
+void BMS_HeatCurrentDischargeLimit(int16_t cellTemp) {
+    if (cellTemp >= 48) {
```
JoshLafleur, line 425: bad logic

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -417,3 +420,25 @@ void BMS_DischargeLimit(uint16_t relativeSoC) {
...
+        BMS.heatChargeCurrentLimit    = -3.75f * cellTemp + 225;
```
JoshLafleur, line 426: im not sure what this formula means

components/bms_worker/include/Environment.h
JoshLafleur, line 1: Remove changes to file
less diffs better

components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -65,6 +65,9 @@ typedef struct
     float       chargeLimit;
     float       dischargeLimit;
 
+    float       heatDischargeCurrentLimit;
+    float       heatChargeCurrentLimit;
+
```
JoshLafleur, line 70: The structure of this is not fully thought out, you now have 2 current limits for each of charge and discharge

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -417,3 +420,25 @@ void BMS_DischargeLimit(uint16_t relativeSoC) {
...
+    } else if (cellTemp > 60) {
+        BMS.heatDischargeCurrentLimit = 0;
+    } else {
+        BMS.heatDischargeCurrentLimit = MAX_CONTINOUS_DISCHARGE_CURRENT;
+    }
+}
+
+void BMS_HeatCurrentChargeLimit(int16_t cellTemp) {
+    if (cellTemp >= 48) {
+        BMS.heatChargeCurrentLimit    = -0.35f * cellTemp + 21;
```
JoshLafleur, line 436: not sure calcs are correct again

JoshLafleur (changes requested):

components/bms_worker/src/BatteryMonitoring.c
tonynokart: i was thinking that it should be 0, if it gets above operating temps. That was just an assumption

components/bms_worker/src/BatteryMonitoring.c
tonynokart: from the range of (48-60°C) it represents a linear function like for the standard ccl and dcl, but with temps on the x-axis. So at 48°C it has a current discharge limit of (-3.75*48+225 = 45A), at 60 (max temp --> -3.75*60+225 = 0A). 
--> 48°C chosen because its 20% from 60, (60*0.8=48)

JoshLafleur:
What i mean by this is that your second if is inclusive of the first. If youre over 60, youre also over 40, so youll never execute the second if. We will cover some core logic functionality together

PR #41, opened 2024-04-02
User/lt/bmsw can float conversions

tonynokart:
I left comments in some areas that bothered me, specifically in regards to the temp and rh variable in HW_HS4011 and HW_SHT40

### DIFF
--- a/components/bms_worker/include/BatteryMonitoring.h
+++ b/components/bms_worker/include/BatteryMonitoring.h
@@ -59,9 +59,9 @@ typedef enum
 
 typedef struct
 {
-    uint16_t   voltage;
-    uint16_t   parasitic_corr;
-    uint16_t   relative_soc;
+    float32_t  voltage;
+    float32_t  parasitic_corr;
+    float32_t  relative_soc;
     BMS_Cell_E state;
 } BMS_Cell_S;
 
@@ -71,25 +71,25 @@ typedef struct
     bool        fault;
     uint16_t    balancing_cells;
     BMS_Cell_S  cells[MAX_CELL_COUNT];      // [mv], precision 1mv
-    uint16_t    pack_voltage;               // [mv], precision 1mv
-    uint16_t    calculated_pack_voltage;    // [mv], precision 1mv
+    float32_t   pack_voltage;               // [mv], precision 1mv
+    float32_t   calculated_pack_voltage;    // [mv], precision 1mv
     uint8_t     connected_cells;
 
-    float charge_limit;
-    float discharge_limit;
+    float32_t charge_limit;
+    float32_t discharge_limit;
 
     struct
     {
-        uint16_t max;
-        uint16_t min;
-        uint16_t avg;
-    } voltage;    // [0.1mv], precision 0.1mv
+        float32_t max;
+        float32_t min;
+        float32_t avg;
+    } voltage;         // [0.1mv], precision 0.1mv
     struct
     {
-        uint16_t min;
-        uint16_t max;
-        uint16_t avg;
-    } relative_soc;    // number from 0-100
+        float32_t min;
+        float32_t max;
+        float32_t avg;
+    } relative_soc;    // [100-0], precision 0.1%
 } BMS_S;
 
 
--- a/components/bms_worker/src/BatteryMonitoring.c
+++ b/components/bms_worker/src/BatteryMonitoring.c
@@ -431,7 +431,7 @@ void BMS_measurementComplete(void)
     {
         for (uint8_t i = 0; i < BMS.connected_cells; i++)
         {
-            BMS.cells[i].parasitic_corr = ((uint32_t)IO.cell[i]) / 128;
+            BMS.cells[i].parasitic_corr = (IO.cell[i]) / 128;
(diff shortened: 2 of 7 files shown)


components/bms_worker/src/Environment.c
```diff
@@ -251,8 +251,8 @@ void ENV_calcTempStats(void)
     uint8_t connected_channels = 0;
 
     ENV.values.avg_temp = 0;
-    ENV.values.max_temp = INT16_MIN;
```
tonynokart, line 254: i set values of -200, as 'impossible' to reach instead of INT16_MIN since max_temp is now a float

components/bms_worker/src/Environment.c
```diff
@@ -162,8 +162,8 @@ static void Environment10Hz_PRD()
         {
             hs_state = DONE;
 
-            ENV.values.board.ambient_temp = hs_chip.data.temp;
-            ENV.values.board.rh           = hs_chip.data.rh;
```
tonynokart, line 166: for hs_chip.data.temp and hs_chip.data.rh , rh and temp are of type int in their respective structs and files. I know it is redundant to divide by 100 or 10 an int because the precision won't improve. I wasn't sure if I should change it in their file HW_HS4011.h/c

components/bms_worker/src/Environment.c
```diff
@@ -176,22 +176,22 @@ static void Environment10Hz_PRD()
     {
         if (SHT_getData())
         {
-            ENV.values.board.ambient_temp = sht_chip.data.temp;
-            ENV.values.board.rh           = sht_chip.data.rh;
```
tonynokart, line 180: same for these two but for the file HW_SHT40.h (only marked in the header file, not in the c file).

components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -71,25 +71,25 @@ typedef struct
...
 
     struct
     {
-        uint16_t max;
-        uint16_t min;
-        uint16_t avg;
-    } voltage;    // [0.1mv], precision 0.1mv
+        float32_t max;
+        float32_t min;
+        float32_t avg;
```
rickyelopez, line 85: unrelated to your changes, but you guys should make a type for this. It's a pretty common pattern.

```c
typedef struct
{
    float32_t max;
    float32_t min;
    float32_t avg;
} MinMaxAvg_S;
``` 
or something to that effect

components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -71,25 +71,25 @@ typedef struct
...
     struct
     {
-        uint16_t min;
-        uint16_t max;
-        uint16_t avg;
-    } relative_soc;    // number from 0-100
+        float32_t min;
+        float32_t max;
+        float32_t avg;
+    } relative_soc;    // [100-0], precision 0.1%
```
rickyelopez, line 92: is this comment actually accurate? What sets the precision to 0.1%?

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -33,11 +33,11 @@
 #endif
 
 #ifndef STANDARD_CHARGE_CURRENT
-# define STANDARD_CHARGE_CURRENT 4.2    // in Amps
+# define STANDARD_CHARGE_CURRENT 4.2f
```
rickyelopez, line 36: Why remove the comment with the unit? These are typically good to have, though I prefer this format:
```suggestion
# define STANDARD_CHARGE_CURRENT 4.2f // [A] Standard charge current
```
Might be better to use the formatting that Doxygen wants

components/bms_worker/src/BatteryMonitoring.c
```diff
@@ -104,7 +104,7 @@ static void BMS1kHz_PRD()
         {
             max_chip.config.sampling_start_100us = UINT32_MAX;
             BMS.state                            = BMS_PARASITIC_MEASUREMENT;
-            BMS.pack_voltage                     = IO.segment * 1000 * 16;
```
rickyelopez, line 107: were these incorrect? or is this change due to something else?

components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -59,9 +59,9 @@ typedef enum
 
 typedef struct
 {
-    uint16_t   voltage;
-    uint16_t   parasitic_corr;
-    uint16_t   relative_soc;
+    float32_t  voltage;
```
rickyelopez, line 62: if you're storing these in an array, be weary of the fact that you've just doubled your flash and/or ram usage significantly by doing this (6 bytes each vs. 12 bytes each)

components/bms_worker/src/Environment.c
rickyelopez: rather than setting some arbitrary float as invalid, you should consider adding a `valid` bit to this struct which you can check separately from the contents of these floats.

components/steering_wheel/lib/lib_FTDI_FT81x
rickyelopez, line 1: what changed here? did you mean to include this submodule?

PR #42, opened 2024-04-28
Created CAN message & some overrun detection

pascalypperciel:
- On teams, it was said that the precision for SOC is 0.01% but in BatteryMonitoring.h it says 0.1%. So I just kind of assumed it really was 0.1%..?
- At the end, with the last two functions (Temperature/Humidity and Fans) I was more liberal with the bits cause there were bits to spare any way. I'm aware that the percentages could've fit in 7 bits lol.

### DIFF
--- a/components/bms_worker/src/CAN/CANIO-tx.c
+++ b/components/bms_worker/src/CAN/CANIO-tx.c
@@ -90,14 +90,10 @@ static const packTable_S MSG_packTable_1Hz[] = {
     { &MSG_pack_BMS_1Hz_Cell_Temp_11_to_19, 0x720, 8U },
     { &MSG_pack_BMS_1Hz_Cell_Voltage_0_to_5, 0x730, 8U },
     { &MSG_pack_BMS_1Hz_Cell_Voltage_6_to_11, 0x740, 8U },
-    { &MSG_pack_BMS_1Hz_Cell_Voltage_12_to_15, 0x750, 8U },
-    { &MSG_pack_BMS_1Hz_Temperatures_and_Humidity, 0x760, 8U },
-    { &MSG_pack_BMS_1Hz_Fans, 0x770, 8U },
+    { &MSG_pack_BMS_1Hz_Cell_Voltage_12_to_15, 0x750, 7U },
+    { &MSG_pack_BMS_1Hz_Temperatures_and_Humidity, 0x760, 5U },
+    { &MSG_pack_BMS_1Hz_Fans, 0x770, 6U },
 };
-
-_Static_assert(MSG_packTable_100Hz_SIZE <= sizeof(MSG_packTable_100Hz) / sizeof(packTable_S), "Size of 100Hz pack table is incorrect.");
-_Static_assert(MSG_packTable_10Hz_SIZE <= sizeof(MSG_packTable_10Hz) / sizeof(packTable_S), "Size of 10Hz pack table is incorrect.");
-_Static_assert(MSG_packTable_1Hz_SIZE <= sizeof(MSG_packTable_1Hz) / sizeof(packTable_S), "Size of 1Hz pack table is incorrect.");
 
 
 /******************************************************************************
@@ -208,12 +204,12 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
 
 static bool MSG_pack_BMS_100Hz_Critical(CAN_data_T* message, const uint8_t counter)
 {
-    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.min / 50) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5115]mV
-    message->u64 |= ((uint64_t)((uint16_t)(BMS.voltage.max / 50) & 0x3FF)) << 44;   //10 bits, 5mV precision, range [0-5115]mV
+    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.min * 200) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5115]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.voltage.max * 200) & 0x3FF)) << 44;   //10 bits, 5mV precision, range [0-5115]mV
     message->u64 |= ((uint64_t)((uint8_t)ENV.values.min_temp & 0x7F)) << 37;        //7 bits, 1 deg C precision
     message->u64 |= ((uint8_t)ENV.values.max_temp & 0x7F) << 30;                    //7 bits, 1 deg C precision
-    message->u64 |= ((uint8_t)BMS.discharge_limit & 0x1F) << 25;                    //5 bits, 1A precision
-    message->u64 |= ((uint8_t)BMS.charge_limit & 0xFF) << 17;                       //8 bits, 1A precision
+    message->u64 |= ((uint8_t)BMS.charge_limit & 0x1F) << 25;                       //5 bits, 1A precision
+    message->u64 |= ((uint8_t)BMS.discharge_limit & 0xFF) << 17;                    //8 bits, 1A precision
     message->u64 |= (((BMS.state == BMS_ERROR) ? 0x01 << 7 : 0U) |
                     ((BMS.fault) ? 0x01 << 6 : 0U) |
                     ((ENV.state == ENV_ERROR) ? 0x01 << 5 : 0U) |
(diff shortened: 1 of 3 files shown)


components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -63,28 +64,38 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
-    { &MSG_pack_BMS_100Hz2, 0x700, 8U },
-    { &MSG_pack_BMS_100Hz3, 0x740, 8U },
+    { &MSG_pack_BMS_1Hz_SOC_Voltage_Temp, 0x700, 8U },
+    { &MSG_pack_BMS_1Hz_Cell_Temp_2_to_10, 0x710, 8U },
+    { &MSG_pack_BMS_1Hz_Cell_Temp_11_to_19, 0x720, 8U },
+    { &MSG_pack_BMS_1Hz_Cell_Voltage_0_to_5, 0x730, 8U },
+    { &MSG_pack_BMS_1Hz_Cell_Voltage_6_to_11, 0x740, 8U },
+    { &MSG_pack_BMS_1Hz_Cell_Voltage_12_to_15, 0x750, 8U },
+    { &MSG_pack_BMS_1Hz_Temperatures_and_Humidity, 0x760, 8U },
+    { &MSG_pack_BMS_1Hz_Fans, 0x770, 8U },
```
JoshLafleur, line 92: Not all of these messages are as small as they can be. if the message uses 48 bits, it should be a 6 byte message

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -63,28 +64,38 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
 };
 
+_Static_assert(MSG_packTable_100Hz_SIZE <= sizeof(MSG_packTable_100Hz) / sizeof(packTable_S), "Size of 100Hz pack table is incorrect.");
+_Static_assert(MSG_packTable_10Hz_SIZE <= sizeof(MSG_packTable_10Hz) / sizeof(packTable_S), "Size of 10Hz pack table is incorrect.");
+_Static_assert(MSG_packTable_1Hz_SIZE <= sizeof(MSG_packTable_1Hz) / sizeof(packTable_S), "Size of 1Hz pack table is incorrect.");
+
```
JoshLafleur, line 98: What is the purpose of these?

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
 
-static bool MSG_pack_BMS_100Hz(CAN_data_T* message, const uint8_t counter)
+static bool MSG_pack_BMS_100Hz_Critical(CAN_data_T* message, const uint8_t counter)
+{
+    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.min / 50) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5115]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.voltage.max / 50) & 0x3FF)) << 44;   //10 bits, 5mV precision, range [0-5115]mV
+    message->u64 |= ((uint64_t)((uint8_t)ENV.values.min_temp & 0x7F)) << 37;        //7 bits, 1 deg C precision
+    message->u64 |= ((uint8_t)ENV.values.max_temp & 0x7F) << 30;                    //7 bits, 1 deg C precision
+    message->u64 |= ((uint8_t)BMS.discharge_limit & 0x1F) << 25;                    //5 bits, 1A precision
+    message->u64 |= ((uint8_t)BMS.charge_limit & 0xFF) << 17;                       //8 bits, 1A precision
```
JoshLafleur, line 213: im pretty sure you will lose bits without uint64_t cast

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
 }
 
-static bool MSG_pack_BMS_100Hz(CAN_data_T* message, const uint8_t counter)
+static bool MSG_pack_BMS_100Hz_Critical(CAN_data_T* message, const uint8_t counter)
+{
+    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.min / 50) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5115]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.voltage.max / 50) & 0x3FF)) << 44;   //10 bits, 5mV precision, range [0-5115]mV
+    message->u64 |= ((uint64_t)((uint8_t)ENV.values.min_temp & 0x7F)) << 37;        //7 bits, 1 deg C precision
+    message->u64 |= ((uint8_t)ENV.values.max_temp & 0x7F) << 30;                    //7 bits, 1 deg C precision
+    message->u64 |= ((uint8_t)BMS.discharge_limit & 0x1F) << 25;                    //5 bits, 1A precision
```
JoshLafleur, line 212: discharge can be 0-255 , charge can be 0-31

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
     return NULL;
 }
 
-static bool MSG_pack_BMS_100Hz(CAN_data_T* message, const uint8_t counter)
+static bool MSG_pack_BMS_100Hz_Critical(CAN_data_T* message, const uint8_t counter)
+{
+    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.min / 50) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5115]mV
```
JoshLafleur, line 208: calculation is wrong. say 4/50 = 0.04 which as an int is 0. you need to work the other way ie voltage *1000/5. *1000 takes volts to mv, and /5 gives it on 10 bit size. also the range is the valid range, and the valid range is 0-5000

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
-    message->u16[1] = BMS.voltage.max;
-    message->u16[2] = BMS.voltage.avg;
-    message->u8[6]  = (BMS.calculated_pack_voltage) / 2;    // Gives range 0-510V
-    message->u8[7]  = (BMS.state == BMS_ERROR) ? 0x01 << 7 : 0U | (BMS.fault)            ? 0x01 << 6
-                                                         : 0U | (ENV.state == ENV_ERROR) ? 0x01 << 5
-                                                         : 0U | (ENV.state == ENV_FAULT) ? 0x01 << 4
-                                                                                         : 0U;
-
-    return true;
+    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.avg / 3) & 0x3FF)) << 54;        //10 bits, 3mv precisiion
```
JoshLafleur, line 232: same problem as above

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
-    message->u16[3] = COOL.rpm[0];
-    message->u16[4] = COOL.rpm[1];
-    return true;
+    message->u64  = ((uint64_t)((uint16_t)(BMS.cells[0].voltage / 5) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5120]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.cells[1].voltage / 5) & 0x3FF)) << 44;   //10 bits, 5mV precision, range [0-5120]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.cells[2].voltage / 5) & 0x3FF)) << 34;   //10 bits, 5mV precision, range [0-5120]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.cells[3].voltage / 5) & 0x3FF)) << 24;   //10 bits, 5mV precision, range [0-5120]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.cells[4].voltage / 5) & 0x3FF)) << 14;   //10 bits, 5mV precision, range [0-5120]mV
+    message->u64 |= ((uint64_t)((uint16_t)(BMS.cells[5].voltage / 5) & 0x3FF)) << 4;    //10 bits, 5mV precision, range [0-5120]mV
+    return true; //60 bits used, 4 unused
```
JoshLafleur, line 282: Same issues as above

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
+}
+
+static bool MSG_pack_BMS_1Hz_Temperatures_and_Humidity(CAN_data_T* message, const uint8_t counter)
+{
+    UNUSED(counter);
+    message->u8[0] = (uint8_t)(ENV.values.board.brd_temp[0] / 10);  //8 bits, 1 deg C precision, range [0-255]deg C
+    message->u8[1] = (uint8_t)(ENV.values.board.brd_temp[1] / 10);  //8 bits, 1 deg C precision, range [0-255]deg C
+    message->u8[2] = (uint8_t)(ENV.values.board.mcu_temp / 10);     //8 bits, 1 deg C precision, range [0-255]deg C
+    message->u8[3] = (uint8_t)(ENV.values.board.ambient_temp / 10); //8 bits, 1 deg C precision, range [0-255]deg C
+    message->u8[4] = (uint8_t)(ENV.values.board.rh / 100);          //8 bits, 1% precision, range [0-255]%
```
JoshLafleur, line 315: these could all be 7 bit 0-127 excet rh which is 0-100

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -202,66 +203,127 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
+    return true; //40 bits used, 24 unused
+}
+
+static bool MSG_pack_BMS_1Hz_Fans(CAN_data_T* message, const uint8_t counter)
+{
+    UNUSED(counter);
+    message->u8[0]  = COOL.percentage[0];   //8 bits, 1% precision, range [0-255]%
+    message->u8[1]  = COOL.percentage[1];   //8 bits, 1% precision, range [0-255]%
```
JoshLafleur, line 323: percentage has range 0-100, can both be 7 bit but 8 bit is simple wont change anything here. arguably can also have 8th bit as fan on or fan off flag for each fan, so like byte 1 has bits 0-6 as percent and bit 7 as on or off

JoshLafleur (changes requested):
Overall pretty good just some misunderstandings throughout the code and few places to improve

components/bms_worker/src/CAN/CANIO-tx.c
pascalypperciel: Actually meant to remove these but forgot to delete the lines, mistake.

PR #43, opened 2024-05-01
UDS App Download

rickyelopez:
Adding support for downloading an app using UDS!

Status:
- [x] Add support for erasing the app flash region
- [x] Add support for SIDs required for app download
  - [x] Add support for starting the download
  - [x] Add support for actually transferring payload and writing to flash
  - [x] Add support for ending the download
- [x] Transfer a full app and boot it
- [x] Add checksums for each block that is transmitted so we can detect failures quickly (and maybe retransmit the block?)
- [ ] Figure out why CRC calculation started before app has finished downloading causes the CPU to hang (maybe we do this later?)
- [x] ~Figure out why ISO-TP is currently only allowing a max block size of 11 bytes per transfer~ idk why but this isn't a problem anymore. I guess one of the other bugs I fixed was causing this too
- [ ] Queue received UDS messages for processing in the 1kHz thread. Right now the last received message is overwritten if a new one comes in, meaning that sending tester present during an app download (for example) will cause a dropped frame

I'm working on building out the UDS Client (which is what folks will run from their computers to actually run UDS operations) [here](https://github.com/concordia-fsae/conUDS), but right now it's incredibly hacky


rickyelopez:
I got a successful app download (and subsequent checksum verification + successful boot), which is pretty sick. There's a bug in the CRC calculation that causes the cpu to hang the crc calculation is started early (before the app has fully downloaded) but I haven't been able to figure out why yet. I circumvented it by adding the UDS boot timeout extend calls (which were required anyway), but we should probably dig into that at some point.

Updated TODOs in the PR description

rickyelopez:
I think this is pretty much ready to go. The two open items in the PR description can happen in a future PR.

Next step here would be to try to integrate the bootloader with actual app (as opposed to the heartbeat I wrote), and to add UDS support to an app (so that we can, at the very least, issue a UDS reset to be able to get into the bootloader and write a new app). I'll probably do both these things with the steering wheel app for now, since that's the board I have on-hand.

@JoshLafleur, wanna take a look?

JoshLafleur (approved):
Not super familiar but looks good to me across everything

PR #44, opened 2024-05-06
New PR, CAN messages

pascalypperciel:
because old one was broken


components/bms_worker/include/BatteryMonitoring.h
```diff
@@ -74,9 +74,9 @@ typedef struct
     BMS_State_E state;
     bool        fault;
     uint16_t    balancing_cells;
-    BMS_Cell_S  cells[MAX_CELL_COUNT];      // [mv], precision 1mv
-    float32_t   pack_voltage;               // [mv], precision 1mv
-    float32_t   calculated_pack_voltage;    // [mv], precision 1mv
+    BMS_Cell_S  cells[MAX_CELL_COUNT];      // [V], precision 1V
```
rickyelopez, line 77: still not clear on where this 1v precision is coming from

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -99,78 +106,68 @@ static const packTable_S MSG_packTable_1Hz[] = {
...
-    if (cantx.tx_100Hz_msg == MSG_packTable_100Hz_SIZE)
+    if (cantx.tx_100Hz_msg != MSG_packTable_100Hz_SIZE)
     {
-        goto continue1;
-    }
-
-    static uint8_t counter_100Hz = 0U;
-    CAN_data_T     message_100Hz;
+        static uint8_t counter_100Hz = 0U;
+        CAN_data_T     message_100Hz;
```
rickyelopez, line 112: this should be initialized to zero
```suggestion
        CAN_data_T     message_100Hz = { 0U };
```

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -99,78 +106,68 @@ static const packTable_S MSG_packTable_1Hz[] = {
...
 
-    // TODO: add overrun detection here
-continue1:
-    if (cantx.tx_10Hz_msg == MSG_packTable_10Hz_SIZE)
+    if (cantx.tx_10Hz_msg != MSG_packTable_10Hz_SIZE)
     {
-        goto continue2;
-    }
+        static uint8_t counter_10Hz = 0U;
+        CAN_data_T     message_10Hz;
```
rickyelopez, line 132: same here

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -99,78 +106,68 @@ static const packTable_S MSG_packTable_1Hz[] = {
...
+    
+    if (cantx.tx_1Hz_msg != MSG_packTable_1Hz_SIZE)
     {
-        return;
-    }
-
-    static uint8_t counter_1Hz = 0U;
-    CAN_data_T     message_1Hz;
+        static uint8_t counter_1Hz = 0U;
+        CAN_data_T     message_1Hz;
```
rickyelopez, line 152: and here

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -205,66 +202,129 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
     return NULL;
 }
 
-static bool MSG_pack_BMS_100Hz(CAN_data_T* message, const uint8_t counter)
+static bool MSG_pack_BMS_100Hz_Critical(CAN_data_T* message, const uint8_t counter)
+{
+    message->u64  = ((uint64_t)((uint16_t)(BMS.voltage.min * 200) & 0x3FF)) << 54;   //10 bits, 5mV precision, range [0-5115]mV
```
rickyelopez, line 207: the way that you've done all these works, however, the more you do this way the more you'll have to **redo** when we implement autogeneration.

See here for how you can do this in such a way that it will easily support conversion to autogeneration:

https://github.com/concordia-fsae/firmware/blob/master/components/steering_wheel/src/CAN/CANIO-tx.c#L69

https://github.com/concordia-fsae/firmware/blob/master/components/steering_wheel/generated/MessagePack.c#L28

https://github.com/concordia-fsae/firmware/blob/master/components/steering_wheel/generated/VEH_sigTx.h#L13

the two in the `generated` dir are the ones that will eventually be autogenerated, and the other one is what we will need to keep doing manually. Note that the autogenerated portion actually encompasses most of the work you've done manually here.

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -283,6 +343,9 @@ static void CANIO_tx_1kHz_PRD(void)
  */
 static void CANIO_tx_100Hz_PRD(void)
 {
+    if (cantx.tx_100Hz_msg < MSG_packTable_100Hz_SIZE) {
```
rickyelopez, line 346: should these check for non-zero?
```suggestion
    if ((cantx.tx_100Hz_msg > 0) && (cantx.tx_100Hz_msg < MSG_packTable_100Hz_SIZE)) {
```
?

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -99,78 +106,68 @@ static const packTable_S MSG_packTable_1Hz[] = {
  */
 void CANTX_BUS_A_SWI(void)
```
rickyelopez, line 107: why is there only one SWI? Doing it this way means that your higher priority messages (100Hz in this case) can be slowed down by your low priority messages (1Hz).

components/bms_worker/include/BatteryMonitoring.h
pascalypperciel: I've been told that when those values have been converted to float32_t, all the values were converted to the "default unit". Maybe I misunderstood though but this is what I've been told.

components/bms_worker/src/CAN/CANIO-tx.c
pascalypperciel: To be discussed with @JoshLafleur

components/bms_worker/src/CAN/CANIO-tx.c
pascalypperciel: Yeah I think it would be better actually.

pascalypperciel: Actually, I don't think this is necessary. From what I remember, those functions are called to reset the counters to zero. I've added those if statements to make sure the counter reached the packTable size (aka all the messages have been sent) before resetting the counter. So I don't really see the point of verifying for non-zeros.

components/bms_worker/src/CAN/CANIO-tx.c
pascalypperciel: I'd love to get on a call with you about this! Will shoot you a message on Teams.

rickyelopez: I'm not on teams, I'm on the fsae discord though if you're on there

components/bms_worker/src/CAN/BMS_msgTx.h
```diff
@@ -0,0 +1,347 @@
...
+
+/******************************************************************************
+ *                     P R I V A T E  F U N C T I O N S
+ ******************************************************************************/
+
+// MSG_pack_BMS_100Hz_Critical
+
+static inline void set_BMSVoltageMin(CAN_data_T* message, uint16_t value) 
+{
+    message->u64  = ((uint64_t)((uint16_t)(value * 200) & 0x3FF)) << 54;  //10 bits, 5mV precision, range [0-5115]mV
```
rickyelopez, line 15: should this be `|=`?

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -66,26 +69,32 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
                                           CAN_data_T*        message,
                                           uint8_t*           nextCounter);
 
-static bool MSG_pack_BMS_100Hz(CAN_data_T* message, const uint8_t counter);
-static bool MSG_pack_BMS_100Hz1(CAN_data_T* message, const uint8_t counter);
-static bool MSG_pack_BMS_100Hz2(CAN_data_T* message, const uint8_t counter);
-static bool MSG_pack_BMS_100Hz3(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_100Hz_Critical(CAN_data_T* message, const uint8_t counter);
```
rickyelopez, line 72: the format for these function names should be as follows:
`pack_BUS_NODE_messageName_period`
i.e. if the message name is BMS_cellData on the VEH bus at 10ms, the name should be`:
`pack_VEH_BMS_cellData_10ms`

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -66,26 +69,32 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
+static bool MSG_pack_BMS_1Hz_SOC_Voltage_Temp(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Cell_Temp_2_to_10(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Cell_Temp_11_to_19(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Cell_Voltage_0_to_5(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Cell_Voltage_6_to_11(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Cell_Voltage_12_to_15(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Temperatures_and_Humidity(CAN_data_T* message, const uint8_t counter);
+static bool MSG_pack_BMS_1Hz_Fans(CAN_data_T* message, const uint8_t counter);
 
 static const packTable_S MSG_packTable_100Hz[] = {
```
rickyelopez, line 83: I realize you're not touching this in your PR, but these pack tables should be named:
`BUS_packTable_period`, eg. `VEH_packTable_10ms`. Fine to leave it as-is here if you don't want to expand the scope, but it should be changed at some point.

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -66,26 +69,32 @@ static const packTable_S* packNextMessage(const packTable_S* packTable,
...
 static const packTable_S MSG_packTable_10Hz[] = {
     { &MSG_pack_BMS_10Hz, 0x10, 8U },
 };
 static const packTable_S MSG_packTable_1Hz[] = {
-    { &MSG_pack_BMS_1Hz, 0x1, 8U },
-    { &MSG_pack_BMS_100Hz, 0x100, 8U },
-    { &MSG_pack_BMS_100Hz1, 0x110, 8U },
-    { &MSG_pack_BMS_100Hz2, 0x700, 8U },
-    { &MSG_pack_BMS_100Hz3, 0x740, 8U },
+    { &MSG_pack_BMS_1Hz_SOC_Voltage_Temp, 0x700, 8U },
```
rickyelopez, line 90: the alignment here seems wrong. Can you run uncrustify on all these? lmk if you don't have my config file

components/bms_worker/src/CAN/CANIO-tx.c
```diff
@@ -99,78 +108,68 @@ static const packTable_S MSG_packTable_1Hz[] = {
...
 void CANTX_BUS_A_SWI(void)
 {
-    // TODO: add overrun detection here
-    if (cantx.tx_100Hz_msg == MSG_packTable_100Hz_SIZE)
+    if (cantx.tx_100Hz_msg != MSG_packTable_100Hz_SIZE)
     {
-        goto continue1;
-    }
+        static uint8_t counter_100Hz = 0U;
+        CAN_data_T     message_100Hz = { 0U };
```
rickyelopez, line 114: do you really need to have a distinct message object for each frequency? You send each message immediately after packing, so I think it should be fine to just reuse the same object but `memset` it to 0 before the next pack function call?

rickyelopez (commented):
lgtm! can you please squash these commits into one commit with a meaningful title/description?

rickyelopez (approved):
you should confirm with Josh that he's OK with you merging into his branch before you actually merge it, but lgtm

rickyelopez:
This got merged as part of the BMSW PR we merged a couple of days ago.

PR #45, opened 2024-05-17
SCons cleanup

rickyelopez:
A bunch of fixes and cleanups that I've been meaning to do forever.


rickyelopez:
@JoshLafleur I suspect this is gonna conflict with your changes. We need to coordinate testing with your branch on top of this one (or vice versa).

Otherwise, I think this is ready?

JoshLafleur (approved):

PR #46, opened 2024-05-17
Pin container version

rickyelopez:
This is super important. Without this, changes to the container that correspond to changes in the firmware repo will make it so that older hashes of the firmware repo can't be built anymore.


JoshLafleur (commented):
Verified it works with my branch

JoshLafleur (approved):

PR #52, opened 2024-06-07
fix --debugger

rickyelopez:


JoshLafleur (approved):
Looks good and tested working

PR #60, opened 2024-09-02
User/jl/cfr24

JoshLafleur:


JoshLafleur:
Mergeable, verified code. A couple issues remaining with CANRX SWI not working with UDS on BMSW and unsigned bit packing is done with a 2's complement number.

PR #85, opened 2024-10-14
Made Nodes platform dependant and not variant dependant. Variants to …

JoshLafleur:
…be used for hardware variants. IE: PCBA ID

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -61,9 +61,8 @@ if platform:
         if targets:
             # TODO: Handle
             pass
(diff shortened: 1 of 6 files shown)


Gray-man (approved):
Looks good to me, however I am not super familiar with SCons so I'd like to better understand the changes made (specifically to the SConscript because that is the most changed file) on a call or something.

JoshLafleur:
**Test Plan**
[(/)] Ensure all nodes compile with appropriate CANTX/RX ID's

PR #87, opened 2024-10-26
User/jl/features bringup

JoshLafleur:


components/bms_boss/HW/mcuConfig.yaml
JoshLafleur, line 1: drop diff

JoshLafleur:
**Test Plan**
- [x] Ensure builds compile with same feature tree
- [ ] Flash bmsw and bmsb node to confirm typical operation
- [ ] Enter HV active
- [ ] Enter run mode
- [ ] Output torque from motor

PR #175, opened 2025-03-23
HIH Driver in BMSB

JoshLafleur:
### Describe changes

1. Generalized the HIH driver from the BMSB
2. Improve the interface and usability of the HIH driver

### Impact

1. Shouldnt have an impact to the functionality, however the driver was in essence re-written

### Test Plan

- [x] Flash BMSB
- [x] Ensure the reported RH and Temperature make sense and are non 0


freeman803:
BMSB sensor is damaged but non-zero valued observed.

PR #195, opened 2025-04-03
Adds new BMSB GPIO config

Gray-man:
### Describe changes

1. Added preprocessor guards to GPIO pin definition and initialization files according to build config IDs.

### Impact

1. Adds new BMSB GPIO configuration for CFR25 platform PCBA ID.
2. Removes setting of IMD_STATUS pin for config ID 1.

### Test Plan

- [ ] Flash BMSB with CFR24 binary (build config 0)
- [ ] Connect the OK_HS pin on the BMSB to 10V, then to 0V to simulate an IMD fault
- [ ] Ensure that the BMSB acknowledges that an IMD fault has occurred


joshlafleur-zip (commented):
Looks pretty shpack. Whatre youre thoughts on reducing the amount of duplicate code in this file with the suggestions?
I didnt go over the pinout doc/schematic, will circle back to that

joshlafleur-zip:
Also I noticed you test plan is not correct. With no HV you could not simulate a isolation failure, and I suspect it is not sensitive enough to detect a low voltage isolation fialure. The IMD is an electrical bridge gap and therefore requires a higher voltage to drive current through the bridge and impact the isolation measurement circuit. TBH for this ticket your not changing any functionality of the IMD, so I would say it is more than sufficient to spoof the OK_HS signal from the IMD (switching the BMSB pin from 0V to 12V and back to 0V) would be sufficient imo

components/bms_boss/HW/HW_gpio_componentSpecific.c
```diff
@@ -148,3 +149,202 @@ const HW_GPIO_S HW_GPIO_pinmux[HW_GPIO_COUNT] = {
         .resetState = HW_GPIO_PINRESET,
     },
 };
+#endif
+
+#if BMSB_CONFIG_ID == 1U
+const HW_GPIO_S HW_GPIO_pinmux[HW_GPIO_COUNT] = {
```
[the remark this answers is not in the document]
Gray-man: I like both of these suggestions & will implement them.

components/bms_boss/src/BMS.c
```diff
@@ -109,6 +109,7 @@ static void BMS100Hz_PRD(void)
         SYS_SFT_closeShutdown();
     }
 
+    #if BMSB_CONFIG_ID == 0U
```
joshlafleur-zip, line 112: defines aligned to left

components/bms_boss/HW/HW_gpio_componentSpecific.c
```diff
@@ -51,41 +59,112 @@ const HW_GPIO_S HW_GPIO_pinmux[HW_GPIO_COUNT] = {
...
-        .pin = GPIO_PIN_8,
+
+    [HW_GPIO_CAN1_RX] = {
         .mode = GPIO_MODE_INPUT,
-        .speed = GPIO_SPEED_FREQ_LOW,
+        .speed = GPIO_SPEED_FREQ_HIGH,
         .pull = GPIO_NOPULL,
         .resetState = HW_GPIO_NOSET,
+
+        #if BMSB_CONFIG_ID == 0U
```
joshlafleur-zip, line 77: Left align #if's

components/bms_boss/HW/include/HW_gpio_componentSpecific.h
```diff
@@ -30,3 +31,35 @@ typedef enum
...
+    HW_GPIO_I2C1_SCL,
+    HW_GPIO_I2C1_SDA,
+    HW_GPIO_IMD_STATUS_MEM,
+    HW_GPIO_BMS_STATUS_MEM,
+    HW_GPIO_CAN2_RX,
+    HW_GPIO_CAN2_TX,
+    HW_GPIO_TSMS_CHG,
+    HW_GPIO_RESET,
+    HW_GPIO_COUNT,
+}HW_GPIO_pinmux_E;
```
joshlafleur-zip, line 63: Keep single def and add to the end of the list

PR #217, opened 2025-04-13
Bringup horn on VCREAR

JoshLafleur:
### Describe changes

1. VCREAR will control the horn based upon vehicle state
2. Bringup vehicle state on VCREAR as follower

### Impact

1. Horn will sound for 1.5s once the vehicle enters `TS_RUN`

### Test Plan

- [x] NA


JoshLafleur (commented):
LGTM, could use function comments in the module but its component specific file and things are more urgent

PR #233, opened 2025-05-25
added adc channel configs

davi4g1:
### Describe changes

Added channels from HW_adc_componentSpecific.h into  HW_adc_componentSpecific.c. Configured these channels according to rank.

### Impact

HW_adc_componentSpecific.c now has all the channels to sample and the ADC will sample in order of rank. 

### Test Plan

- Input known signals into the analog inputs of VCFront and monitor the values read by ADC.


JoshLafleur (commented):
I didnt review the channel mapping on the schematic or if its inline with the component specific enum definition. While youre here though, copy the changes over to vcrear as well because itll have the same issue. Did you double check the ordering and the mapping? If so ill be good to approve and ship it

PR #234, opened 2025-05-26
fixed variants.yaml bug

davi4g1:
### Describe changes

1. Fixed app_component_id: to the correct component in variants.yaml in /bootloaders

### Impact

1. hwDesc->appComponentId == appDesc->appComponentId; line 60, of LIB_app.c returns true. Bootloader now runs.

### Test Plan

- Run bootloader. This has been done.


JoshLafleur (approved):
This also fixes the vc bootloader led issue

PR #250, opened 2025-06-08
Fix ADC unpacking in VC

JoshLafleur:
### Describe changes

ADC unpacking was still driven by the DMA buffer. This caused an unnecessary unpacking of the buffer, as well as race conditions within the memory which could significantly impact the way that "simultaneous measurements" were taken by the firmware. This change improves the interface and allows the application to unpack the buffer if it wants to manually.

### Impact

1. This will fix the "simultaneous" measurement read by the applications and ensure they are measured properly at the same time
2. Propogate all fixes to BMSB, SWS, VCFRONT, REAR, and PDU - This does not diff BMSW which will take some more figuring out with the 10kHz + normal codepath
### Test Plan

- [x] The original measurement methodology impacted how the apps were reading the measured voltages. Make sure the apps are properly reading the in sync values


JoshLafleur (commented):
lgtm

Issue #261, opened 2025-06-11
5V HSD's not turning on in VCFRONT and VCREAR


davi4g1:
VCFRONT and VCREAR 5V HSD's not turning on. Simple fix is to flip the active state in shared/code/DRV/drv_tps20xx.c line 241.

PR #262, opened 2025-06-11
Flipped active state of HSD in shared/drv_tps20xx.c line 241. Issue #261

davi4g1:
### Describe changes

1. Flipped the active state of the HSD tps20xx in shared/code. This is because the tsp20xx used in VCFRONT and VCREAR is NEN.

### Impact

1. HSD's now turn on in VCFRONT and VCREAR

### Test Plan

- [x] Probe 5V out voltage on VCU's

** This may not be the best or most permanent fix since it's in shared code but comp is in 4 days

### DIFF
--- a/components/shared/code/DRV/drv_tps20xx.c
+++ b/components/shared/code/DRV/drv_tps20xx.c
@@ -238,7 +238,7 @@ void drv_tps20xx_setEnabled(drv_tps20xx_channel_E channel, bool enabled)
  */
 static void drv_tps20xx_private_setICEnabled(drv_tps20xx_channel_E channel, bool enabled)
 {
-    const drv_io_activeState_E state_to_set = (enabled) ? DRV_IO_ACTIVE : DRV_IO_INACTIVE;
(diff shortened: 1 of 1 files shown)


components/shared/code/DRV/drv_tps20xx.c
```diff
@@ -238,7 +238,7 @@ void drv_tps20xx_setEnabled(drv_tps20xx_channel_E channel, bool enabled)
  */
 static void drv_tps20xx_private_setICEnabled(drv_tps20xx_channel_E channel, bool enabled)
 {
-    const drv_io_activeState_E state_to_set = (enabled) ? DRV_IO_ACTIVE : DRV_IO_INACTIVE;
+    const drv_io_activeState_E state_to_set = (enabled) ? DRV_IO_INACTIVE : DRV_IO_ACTIVE;
```
JoshLafleur, line 241: This is just adding bad vibes into the codebase and not fixing a problem which will fuck us in other places

JoshLafleur (commented):
Please debug why setting this active is turning it off. Its a problem in outputAD or in the pin configurations imo

JoshLafleur:
Problem can also be the reading of the fault pin/bad logic in the periodic once its turned on

PR #274, opened 2025-06-13
Fix GPIO output config of VCPDU so HSD's dont auto-retry

JoshLafleur:
### Describe changes

1. Currently the HSDs all auto-retry on a fault. This should not occur until the application is ready to reset

### Impact

1. This will prevent the HSDs from auto-resetting

### Test Plan

- [ ] Flash VCPDU and ensure the HSDs all turn on


JoshLafleur:
Still doesnt work but pins set properly lol, wack

PR #296, opened 2025-07-02
Steering Angle Sensor Bringup

Louiscars190:
### Describe changes

1. Describe changes...
- Added steeringAngle.c
- Added steeringAngle.h
- Changed ADC Spare2 to STR_ANGLE
- Added to VC Front CAN 

### Impact

1. Describe impact...
- Implements steering angle sensor 
- Broadcasts steering angle over CANbus

### Test Plan

- [x] Flash to VCFront
- [x] Connect Candapter and use SavvyCAN to check if the message shows up
- [x] Turn the steering wheel left and right to see if the angle registers


JoshLafleur (dismissed):
Looks great, you can merge if you want or fix and ill review right away, otherwise you can make some improvements

components/vc/front/src/steeringAngle.c
```diff
@@ -0,0 +1,86 @@
...
+
+static lib_interpolation_point_S steering_angle[] = {
+    {
+        .x = 0, // sensor reference voltage
+        .y = -172.5f, // left turned degrees
+    },
+    {
+        .x = 1.5, // dead center voltage
+        .y = 0.0f, // 0 degrees centered
+    },
```
JoshLafleur, line 36: This can be skipped but its fine. Its linear across the entire inout dataset

network/definition/data/components/vcfront/vcfront-signals.yaml
```diff
@@ -47,6 +47,17 @@ signals:
     description: Status of the run button switch
     discreteValues: digitalStatus
 
+  steeringAngle:
+        description: Steering Angle Sensor
+        unit: 'deg'
+        nativeRepresentation:
+          resolution: 1
```
JoshLafleur, line 54: Didnt notice before but this is optional, you can delete this line and itll calculate the smallest resolution you can do on your 9 bits

JoshLafleur (approved):
Looks good if you want to merge, but maybe changes and then squash all commits into just 1? Itll be a cleaner log that way

network/definition/data/components/vcfront/vcfront-signals.yaml
Louiscars190: Oh sick

components/vc/front/src/steeringAngle.c
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * @brief 100ms (10Hz) Function that periodically calls drv_inputAD_getAnalogVoltage
+ * @note Uses lib_interpolation library to map and interpolate values, takes into account edge cases with saturation
+ * @note Has a voltage divider compensation of 1.681 for 3v to 5v
+ */
+static void steeringAngle_periodic_10Hz(void)
+{
+    steeringAngle_data.voltage = 1.681f * drv_inputAD_getAnalogVoltage(DRV_INPUTAD_ANALOG_STR_ANGLE);
```
JoshLafleur, line 71: You mapped relative to 0-3 V, but you convert to 0-5V here. Was this intended?

components/vc/rear/include/HW/drv_inputAD_componentSpecific.h
```diff
@@ -28,7 +28,7 @@ typedef enum
     DRV_INPUTAD_ANALOG_PU1,
     DRV_INPUTAD_ANALOG_BR_POT,
     DRV_INPUTAD_ANALOG_SPARE1,
-    DRV_INPUTAD_ANALOG_SPARE2, 
+    DRV_INPUTAD_ANALOG_STR_ANGLE,
```
JoshLafleur, line 31: We dont use steering sensor on rear, leave this undiffed

JoshLafleur (commented):
Looks good although caught a logic issue you should have seen in testing.

Also the commit message should be
Title description of change (something like bringup steering angle sensor on vcfront)

Description (whatever you want to note)

What you have there is not clear when reviewing logs, and you can 'git commit --amend' and rewrite that. Also you should test your code to make sure it works and when done check off your test cases

components/vc/rear/include/HW/drv_inputAD_componentSpecific.h
Louiscars190: Ok, I'll change

components/vc/front/src/steeringAngle.c
```diff
@@ -0,0 +1,89 @@
...
+        .x = 0.345f, // voltage
+        .y = 102.0f, // right turned degrees
+    },  
+};
+
+static lib_interpolation_mapping_S steering_map = {
+    .points = (lib_interpolation_point_S*)&steering_angle,
+    .number_points = COUNTOF(steering_angle),
+    .saturate_left = false,
+    .saturate_right = true,
```
JoshLafleur, line 47: Why are you only saturating in one direction? Shouldnt it be both if theyre both at the limit?

components/vc/front/src/steeringAngle.c
```diff
@@ -0,0 +1,89 @@
...
+
+static lib_interpolation_point_S steering_angle[] = {
+    {
+        .x = 1.414f, // sensor reference voltage
+        .y = -102.0f, // left turned degrees
+    },
+    {
+        .x = 0.894f, // voltage
+        .y = 0.0f, // center degrees
+    },
```
JoshLafleur, line 36: Why is this not exactly between the two other points? Is the scale of the sensor different when on one half versus the other half?

Louiscars190: This is the only way we could get it to be linear we tried a bunch of different things. The sensor goes from 0-3 then once its passed a threshold it goes back to zero spinning continuously

components/vc/front/src/steeringAngle.c
joshlafleur-zip: saturation shouldnt impact linearity. that souds jank af imo but whatever if it works who cares about tech debt in formula

components/vc/front/src/steeringAngle.c
joshlafleur-zip: hmmm maybe ill take a look when im back in town then. remind me

PR #300, opened 2025-08-12
User/kz/vc brake temp

kelvinzigah:
### Describe changes

1. Added brakeTemp.h
2. Added brakeTemp.c

### Impact

1. Reads brake rotor temperature
2. Converts it into voltage
3. outputs both over CAN

### Test Plan

- [ ]  Flash VCFRONT/VCREARs
- [ ] Use SavvyCAN to check if the car outputs message
- [ ] Tune the sensors to make sure the messages sent over CAN are accurate


JoshLafleur:
> I think this is largely fine with the exception of the commnts I've made. Double check with @JoshLafleur before making any of the changes I've suggested though.
> 
> 
> 
> One big thing that I think you should do differently here: instead of re-defining basically exactly the same file, once for VCFRONT and once for VCREAR, you should build this file once in a shared directory and then use `#if`s and feature flags to make everything generic. Then, you can auto-generate a header file that `#define`s everything that you need in `brakeTemps.c`.
> 
> 
> 
> Sharing code like this minimizes the surface area for bugs to be introduced, and eliminates the possibility of forgetting to update one of the copies in the future if you find a bug.

Suggestions look good (i didnt go deep through and review), I did the stupid way setting up the VCs but yes sharing code would be better across both, we can make a 'vc/shared' folder easily across both. No codegen setup for the structs but thats a pretty easy one to bang through, but I wouldnt say is necessary. For interpolation library, saturation is done to clip the values (your comment was correct). @kelvinzigah you need to have the mapping in ascending X, thats why you have the issue with saturation on one side

PR #305, opened 2025-08-14
bmsb: Add pack current sense offset

JoshLafleur:
### Describe changes

1. Calibrate away 0 A pack current sense

### Impact

1. Pack reports ~0 Amps when sitting

### Test Plan

- [ ] Flash BMSB, ensure ~0 A

### DIFF
--- a/network/definition/data/components/bmsb/bmsb-signals.yaml
+++ b/network/definition/data/components/bmsb/bmsb-signals.yaml
@@ -89,6 +89,16 @@ signals:
         max: 400
     continuous: true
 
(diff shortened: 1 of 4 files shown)


network/definition/data/components/bmsb/bmsb-signals.yaml
```diff
@@ -89,6 +89,16 @@ signals:
         max: 400
     continuous: true
 
+  packCSVoltage:
```
rickyelopez, line 92: `packCurrentSenseV`?

PR #310, opened 2025-08-16
added messages to bridge

freeman803:
### added messages

to go from ass bus to veh bus

### Test Plan

- [ ] flash vcrear
- [ ] see if messages show up on both busses

### DIFF
--- a/network/definition/data/components/vcrear/vcrear-rx.yaml
+++ b/network/definition/data/components/vcrear/vcrear-rx.yaml
@@ -12,6 +12,19 @@ messages:
   BMSB_currentLimit:
     destBuses: ass
     sourceBuses: veh
+  PM100DX_temps1:
+    destBuses: veh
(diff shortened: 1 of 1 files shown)


JoshLafleur (approved):

network/definition/data/components/vcrear/vcrear-rx.yaml
```diff
@@ -12,6 +12,19 @@ messages:
   BMSB_currentLimit:
     destBuses: ass
     sourceBuses: veh
+  PM100DX_temps1:
+    destBuses: veh
+    sourceBuses: ass
```
rickyelopez, line 17: @JoshLafleur why does this need to be specified in the rx file? We should be able to infer this on the rx side, unless there are multiple possible paths (in which case we should pick the shortest one by default unless manually specified)

PR #311, opened 2025-08-17
wheelSpeed: Implement vcfront and vcrear

JoshLafleur:
### Describe changes

Bringup wheelspeeds on vcfront and vcrear. Share TIM firmware between vcfront and vcrear

### Impact

1. Describe impact...

### Test Plan

- [x] Ensure rear wheel speeds are 4.6x less than motor speed
- [x] Ensure right wheels increases right wheel speed signal


components/vc/front/SConscript
```diff
@@ -158,6 +160,7 @@ project_source_files = [
     SRC_DIR.File("brakePressure.c"),
     SRC_DIR.File("steeringAngle.c"),
     SRC_DIR.File("shockpot.c"),
+    SHARED_VC.File("wheelSpeed.c"),
```
rickyelopez, line 163: keeping these alphabetized will help if we switch to buck

network/definition/data/components/vcfront/vcfront-message.yaml
```diff
@@ -74,3 +74,17 @@ messages:
...
+    cycleTimeMs: 50
+    signals:
+      velocityL:
+        template: tireLinearVelocity
+      velocityR:
+        template: tireLinearVelocity
+      rpmL:
+        template: tireRPM
+      rpmR:
+        template: tireRPM
```
rickyelopez, line 90: Do you actually need near velocity for something?

IMO these signal names aren't great. `VCFRONT_velocityL` and such... Maybe `VCFRONT_wheelVelocityL` and `VCFRONT_wheelSpeedL` or `VCFRONT_wheelRpmL`?

network/definition/data/templates/signals.yaml
```diff
@@ -95,3 +95,21 @@ signals:
         min: 0.0
         max: 15
     continuous: true
+  tireLinearVelocity:
```
rickyelopez, line 98: the c code all calls it wheel speed but here you're effectively calling it tire speed, maybe pick one and stick to it?

JoshLafleur:
@rickyelopez this is a shamefull first attempt at sharing it im cleaning it up this evening. Should be a draft atp

JoshLafleur:
> > @rickyelopez this is a shamefull first attempt at sharing it im cleaning it up this evening. Should be a draft atp
> 
> idk if I'd call it shameful lmao but kk sg

Ready for review. If we want to alphabetize the sources we can spin up a new ticket for a new member to learn scons a bit

components/vc/front/SConscript
rickyelopez: actually I take it back. alphabetization doesn't matter, we just need to preserve the order when we re-implement this in buck. We will for sure have to re-order some stuff in scons in order to make it hex-neutral vs. buck though, but we can do that later

network/definition/data/components/vcrear/vcrear-message.yaml
```diff
@@ -76,3 +76,17 @@ messages:
...
+
+  wheelSpeed:
+    description: Rear wheel speeds
+    id: 0x460
+    cycleTimeMs: 50
+    signals:
+      wheelVelocityL:
+        template: wheelLinearVelocity
+      wheelVelocityR:
+        template: wheelLinearVelocity
```
rickyelopez, line 88: Based on your templates, these aren't actually velocities, they're magnitudes.
I'd go with
```suggestion
      wheelSpeedLinearL:
        template: wheelLinearVelocity
      wheelSpeedLinearR:
        template: wheelLinearVelocity
```

network/definition/data/components/vcrear/vcrear-message.yaml
```diff
@@ -76,3 +76,17 @@ messages:
...
+      wheelRpmL:
+        template: wheelRPM
+      wheelRpmR:
+        template: wheelRPM
```
rickyelopez, line 92: rpm is a unit and probably shouldn't be in the signal name (unless you have two signals that are otherwise identical other than the unit, which is not the case here). Also, I know that vcfront publishes fronts and vcrear publishes rears, but it might be nice to have F/R in here regardless. I think it would be more obvious if you used
```suggestion
      wheelSpeedRL:
        template: wheelRPM
      wheelSpeedRR:
        template: wheelRPM
```
and same for fronts

network/definition/data/templates/signals.yaml
```diff
@@ -95,3 +95,21 @@ signals:
...
         max: 15
     continuous: true
+  wheelLinearVelocity:
+    unit: 'm/s'
+    nativeRepresentation:
+      bitWidth: 10
+      resolution: 0.01
+      range:
+        min: 0
+        max: 50
```
rickyelopez, line 105: Do you need to specify all three of `bitWidth`, `resolution`, and `range.max`? I think yamcan can auto-calculate the most optimal settings if you omit some of these. Namely, in this case, probably fine to use just `range` and `resolution`

network/definition/data/templates/signals.yaml
```diff
@@ -95,3 +95,21 @@ signals:
...
+    continuous: true
+  wheelRPM:
+    unit: 'RPM'
+    nativeRepresentation:
+      bitWidth: 12
+      resolution: 1
+      range:
+        min: 0
+        max: 2000
```
rickyelopez, line 114: same here. Probably jsut use `range` and `resolution`

components/vc/shared/code/HW_tim_vcSpecific.c
```diff
@@ -0,0 +1,114 @@
...
+/******************************************************************************
+ *                       P U B L I C  F U N C T I O N S
+ ******************************************************************************/
+
+/**
+ * @brief  Initializes TIM peripherals
+ *
+ * @retval true = Success, false = Failure
+ */
+HW_StatusTypeDef_E HW_TIM_init(void)
```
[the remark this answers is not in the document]
rickyelopez: Well, right now HW abstracts between the app and the hardware for one particular hardware line. If you ever decided you wanted to run this code on a non-ST chip or even an ST chip with a different HAL, this breaks down pretty quickly.

Ideally what you want is effectively an internal library that defines stuff like a `timerInit` function where we can use our own internally defined data structures and configuration parameters. Then, in the background, that becomes specialized to the particular HAL that is currently in use.

components/vc/shared/code/HW_tim_vcSpecific.c
```diff
@@ -0,0 +1,144 @@
...
+{
+    switch (tim->Channel)
+    {
+        case HAL_TIM_ACTIVE_CHANNEL_3:
+            for (uint8_t i = 1; i < NUM_TIM_SAMPLES; i++)
+            {
+                wheelSpeed.lastSample[HW_TIM_CHANNEL_WS_R][i - 1U] = wheelSpeed.lastSample[HW_TIM_CHANNEL_WS_R][i];
+            }
+            wheelSpeed.lastSample[HW_TIM_CHANNEL_WS_R][CURRENT_SAMPLE] = HW_TIM_getBaseTick();
+            break;
```
rickyelopez, line 126: this is fine for a small number of samples like this, but it may be worth investing in a circular buffer library in the future

components/vc/front/SConscript
```diff
@@ -27,6 +27,7 @@ SHARED_APP = SHARED_CODE.Dir("app")
 SHARED_HW = SHARED_CODE.Dir("HW")
 SHARED_RTOS = SHARED_CODE.Dir("RTOS")
 SHARED_DRV = SHARED_CODE.Dir("DRV")
+SHARED_VC = Dir("../shared/code/")
```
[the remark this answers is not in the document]
rickyelopez: if the effort is declining now, imagine how unlikely it is that you'll come back and fix this later. Better to try to set it up in a way that it'll work out nicely in the long term.

components/vc/front/SConscript
```diff
@@ -170,7 +173,7 @@ project_hw_files = [
     HW_DIR.File("HW_intc.c"),
     HW_DIR.File("HW_msp.c"),
     SHARED_HW.File("HW_tim.c"),
-    HW_DIR.File("HW_tim_componentSpecific.c"),
+    SHARED_VC.File("HW_tim_vcSpecific.c"),
```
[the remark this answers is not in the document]
rickyelopez: IMO the only reasons to have suffixes like this are:
1. you have to because otherwise two files in the same directory would collide (i.e. `HW_tim.c` and `HW_tim_componentSpecific.c` both need to be in the same directory)
2. it's a header and you may need to import both of them in the same translation unit, so you need them to have different names

Otherwise, you should lean on the directory structure for this kind of inference. This keeps the total path length shorter and reduces the amount of redundant information you have in the path, which, for example, makes it much easier to scroll through a long list of search results.

components/vc/shared/code/HW_tim_vcSpecific.c
```diff
@@ -0,0 +1,144 @@
...
+ *                         P R I V A T E  V A R S
+ ******************************************************************************/
+
+TIM_HandleTypeDef htim[HW_TIM_PORT_COUNT];
+typedef struct
+{
+    uint64_t lastSample[HW_TIM_CHANNEL_WS_CNT][NUM_TIM_SAMPLES];
+} wheelSpeed_data_S;
+
+static wheelSpeed_data_S wheelSpeed;
```
rickyelopez, line 37: god the body controls coding is so strong in you lmao, this is the kind of stuff I don't like about how body controls writes their code

this is a private variable, it will only ever be accessed within the context of this file. `wheelSpeed_data_S` is not any more clear than `ws_S` when you're in the context of this file, but it is longer and more annoying. It's not like you have to jump to a different file to see what it means, plus with modern IDE completion you get it for free with `Shift+k` or hover. Same for `wheelSpeed` vs just `WS`.

components/vc/shared/code/wheelSpeed.h
```diff
@@ -0,0 +1,46 @@
...
+/******************************************************************************
+ *                             T Y P E D E F S
+ ******************************************************************************/
+
+typedef enum
+{
+    WHEELSPEED_SENSOR_L = 0x00,
+    WHEELSPEED_SENSOR_R,
+    WHEELSPEED_SENSOR_CNT,
+} wheelSpeed_Sensor_E;
```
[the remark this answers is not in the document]
JoshLafleur: updated to generic wheel

components/vc/shared/code/HW_tim_vcSpecific.c
JoshLafleur: yea circular buffer library would help here, i agree its not the cleanest solution right now

rickyelopez:
> @rickyelopez Rickster respeck your elders

chat what does he mean by this

PR #323, opened 2025-09-03
User/yto/brake bias

yanntotchoun:
1. Describe changes...
Used a generated getter function to access the rear brake pressure inside of the front brake pressure file. Used it to calculate the brake bias. Output it on the steering wheel by broadcasting a message on CAN.

1. Describe impact...
With this change, we will be able to get a more accurate brake bias because doing math on the steering wheel isn't viable since it can only use integers so it really isn't accurate.

We built the code and we used SavvyCan code to see the CAN message. Then we verified that our calculations were correct by testing the brakes by making sure that our brake bias was different then the one on the steering wheel, but still realistic. We also verified that our getter function was accurate by outputting it on CAN and verifying with the real brake pressure.


rickyelopez (commented):
You need to run uncrustify on you changed C files here and also squash all of your commits into one commit with a solid commit message and description.

components/vc/front/src/brakePressure.c
```diff
@@ -22,8 +22,16 @@ static struct
 {
     float32_t voltage;
     float32_t pressure;
+    float32_t bias;
+    float32_t rearbrpr;
+
 } brakePressure_data;
 
+
```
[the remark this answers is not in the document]
yanntotchoun: ok thank you !

components/vc/front/src/brakePressure.c
```diff
@@ -22,8 +22,16 @@ static struct
 {
     float32_t voltage;
     float32_t pressure;
+    float32_t bias;
+    float32_t pressure_rear;
```
rickyelopez, line 26: we use camel case for variables like this, so should be `pressureRear`

components/vc/front/src/brakePressure.c
```diff
@@ -54,12 +73,17 @@ static void brakePressure_periodic_100Hz(void){
...
+}
 
 /******************************************************************************
  *                           P U B L I C  V A R S
  ******************************************************************************/
 
 const ModuleDesc_S brakePressure_desc = {
     .moduleInit = &brakePressure_init,
-    .periodic100Hz_CLK = &brakePressure_periodic_100Hz,
+    .periodic100Hz_CLK = &brakePressureModule_periodic_100Hz,
```
rickyelopez, line 88: you should not do this. Leave this as is, add the call to `calcBrakeBias_100Hz` to the existing `brakePressure_periodic_100Hz` function.

If you really want to, you can move the code that is already in `brakePressure_periodic_100Hz` to a new function called `calcBrakePressure_100Hz` and call that function from the periodic as well.

components/vc/front/src/brakePressure.c
```diff
@@ -32,10 +40,21 @@ float32_t brakePressure_getBrakePressure(void){
...
+float32_t brakePressure_getBrakeBias(void){ 
+    return brakePressure_data.bias;
+}
+
 static void brakePressure_init(void)
 {
     memset(&brakePressure_data, 0x00U, sizeof(brakePressure_data));
 }
+   
+static void brakeBias_periodic_100Hz(void){
```
rickyelopez, line 52:
```suggestion
static void calcBrakeBias_100Hz(void){
```

components/vc/front/src/brakePressure.c
yanntotchoun: I’ll update it as suggested. Out of curiosity, could you share why we prefer this approach over using a wrapper?

rickyelopez (approved):
lgtm with the exception of the newline change in one of the yaml files! Make sure @JoshLafleur is on-board before you merge, but seems fine to me.

rickyelopez (changes requested):
oh wait actually please squash your commits before merging this! lmk if you need help to do this

components/vc/front/src/brakePressure.c
yanntotchoun: ok thank you

components/vc/front/src/brakePressure.c
```diff
@@ -22,44 +22,60 @@ static struct
...
 {
     memset(&brakePressure_data, 0x00U, sizeof(brakePressure_data));
 }
 
-static void brakePressure_periodic_100Hz(void){
+static void calcBrakeBias_100Hz(void)
+{
+    CANRX_VEH_get_VCREAR_brakePressure(&brakePressure_data.pressureRear);
+    brakePressure_data.bias = (brakePressure_data.pressure / (brakePressure_data.pressure + brakePressure_data.pressureRear)) * 100;
+}
```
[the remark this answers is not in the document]
rickyelopez: eh, the compiler will almost certainly inline it and it's a bit cleaner. Not strictly necessary, but IMO it's fine

yanntotchoun:
> oh wait actually please squash your commits before merging this! lmk if you need help to do this

I wanted to wait until I retested the code until I merge, should I do it right now?

rickyelopez:
> > oh wait actually please squash your commits before merging this! lmk if you need help to do this
> 
> I wanted to wait until I retested the code until I merge, should I do it right now?

Ehhhh it's kind of up to you, as long as it's done before merging.

_technically_ if you want to do it properly you should only be merging the exact commit hash that you tested, which means you should do your testing on a squashed commit. It's pretty easy to just keep everything in one commit most of the time using `git commit --amend` but it does make it a little harder to review sometimes since you can't click the "changes since I last reviewed" button

components/vc/front/src/brakePressure.c
```diff
@@ -22,44 +22,60 @@ static struct
...
 static void brakePressure_init(void)
 {
     memset(&brakePressure_data, 0x00U, sizeof(brakePressure_data));
 }
 
-static void brakePressure_periodic_100Hz(void){
+static void calcBrakeBias_100Hz(void)
+{
+    CANRX_VEH_get_VCREAR_brakePressure(&brakePressure_data.pressureRear);
+    brakePressure_data.bias = (brakePressure_data.pressure / (brakePressure_data.pressure + brakePressure_data.pressureRear)) * 100;
```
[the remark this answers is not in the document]
rickyelopez: saturating is when you clamp a value between a minimum and a maximum, so like if you saturate the number 1.5 between 0 and 1 you get 1. The path Josh shared has a `SATURATE` macro that you could use here that would ensure that this math doesn't somehow break and give you like 100000 or something lol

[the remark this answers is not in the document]
yanntotchoun: ok thank you !

yanntotchoun:
> > > oh wait actually please squash your commits before merging this! lmk if you need help to do this
> > 
> > 
> > I wanted to wait until I retested the code until I merge, should I do it right now?
> 
> Ehhhh it's kind of up to you, as long as it's done before merging.
> 
> _technically_ if you want to do it properly you should only be merging the exact commit hash that you tested, which means you should do your testing on a squashed commit. It's pretty easy to just keep everything in one commit most of the time using `git commit --amend` but it does make it a little harder to review sometimes since you can't click the "changes since I last reviewed" button

what would be the commands that I would do to squash all of the 21 commits

PR #329, opened 2025-09-06
yamcan: Add buck2 build support

JoshLafleur:
### Describe changes

1. Describe changes...

### Impact

1. Describe impact...

### Test Plan


components/bms_boss/BUCK
```diff
@@ -0,0 +1,355 @@
...
+        "queue.h": "//embedded/libs:FreeRTOS.git[Source/include/queue.h]",
+        "stack_macros.h": "//embedded/libs:FreeRTOS.git[Source/include/stack_macros.h]",
+        "stream_buffer.h": "//embedded/libs:FreeRTOS.git[Source/include/stream_buffer.h]",
+        "task.h": "//embedded/libs:FreeRTOS.git[Source/include/task.h]",
+        "timers.h": "//embedded/libs:FreeRTOS.git[Source/include/timers.h]",
+    },
+    headers = {
+        "FreeRTOSConfig.h": "RTOS/include/FreeRTOSConfig.h",
+    },
+)
```
rickyelopez, line 119: we should commonize this like I did for the HAL

components/steering_wheel/BUCK
```diff
@@ -146,7 +146,7 @@ stm32f1_hal(
 cxx_genrule(
     name = "linkscript",
     _cxx_toolchain = TOOLCHAIN,
-    cmd = "$(cpp) -P -undef {}".format(" ".join(compiler_flags)) + " ${SRCS} | sed -e '/^#.\\+$/d' > ${OUT}",
+    cmd = "set -o pipefail && $(cpp) -P -undef {}".format(" ".join(compiler_flags)) + " ${SRCS} | sed -e '/^#.\\+$/d' > ${OUT}",
```
rickyelopez, line 149: does this actually work? I'd expect this to have to be `set -o pipefail; ...`

components/bms_boss/BUCK
```diff
@@ -0,0 +1,355 @@
...
+
+export_file(
+    name = "FeatureDefines_generated.h",
+    src = "generated/FeatureDefines_generated.h"
+)
+
+export_file(
+    name = "BuildDefines_generated.h",
+    src = "generated/BuildDefines_generated.h"
+)
```
rickyelopez, line 40: why?

network/NetworkGen/requirements.txt
rickyelopez, line 1: do you need this?

components/shared/code/defs.bzl
```diff
@@ -0,0 +1,26 @@
+load("@prelude//:rules.bzl", __rules__ = "rules")
+
+def strip_prefix(s, p):
+    return s[len(p):] if s.startswith(p) else s
+
+def remap_headers(bases):
```
rickyelopez, line 6: feels like this function should also take in the list of headers instead of doing the glob internally

components/shared/code/BUCK
```diff
@@ -0,0 +1,24 @@
...
+)
+
+folders = [
+    "DRV/",
+    "HW/",
+    "RTOS/",
+    "app/",
+    "libs/",
+]
+exported_headers = remap_headers(folders)
```
rickyelopez, line 16: per comment in defs.bzl, you would do the glob out here and pass in the list to `remap_headers`. Then this becomes a general utility function that you could move somewhere else instead of being specific to this dir

components/shared/code/BUCK
```diff
@@ -0,0 +1,24 @@
+load("//components/shared/code/defs.bzl", "remap_headers")
+
+filegroup(
```
rickyelopez, line 3: is this really what you want? one `srcs` that contains all your shared code?

components/bms_boss/BUCK
```diff
@@ -0,0 +1,355 @@
...
+    header_only = True,
+    header_namespace = "",
+    exported_headers = {
+        "BuildDefines.h": ":BuildDefines.h",
+        "BuildDefines_generated.h": ":BuildDefines_generated.h",
+        "FeatureDefines_generated.h": ":FeatureDefines_generated.h",
+    }
+)
+
+compiler_flags += ["-include", "BuildDefines.h"]
```
rickyelopez, line 133: why is this necessary? what c file uses this header but doesn't include it?

components/bms_boss/BUCK
```diff
@@ -0,0 +1,355 @@
...
+
+shared_code_library(
```
rickyelopez, line 135: this compiles everything in `shared/code`, right? That seems weird to me...

components/bms_boss/BUCK
```diff
@@ -0,0 +1,355 @@
...
+        cmd = "set -o pipefail && $(cpp) -P -undef {}".format(" ".join(compiler_flags)) +
+            " -include BuildDefines.h -include FeatureDefines_generated.h linkscript | sed -e '/^#.\\+$/d' > ${OUT}",
+        out = "linkscript",
+        srcs = {
+            "linkscript": "//embedded/platforms/stm32/f1/" + variant + ":" + script,
+            "BuildDefines.h": ":BuildDefines.h",
+            "BuildDefines_generated.h": ":BuildDefines_generated.h",
+            "FeatureDefines_generated.h": ":FeatureDefines_generated.h",
+        },
+    )
```
rickyelopez, line 254: yeah this defs needs to become a proper macro call lol, this is gonna become janky so fast. you don't have to do it here, just calling it out

components/bms_boss/BUCK
joshlafleur-zip: Temporary solution until feature tree is built out

components/shared/code/defs.bzl
joshlafleur-zip: It globs relative to the caller so it saves code in there. You really provide it a list of sources relative to the caller and itll output a mapping of stripped: full file location. Should be moved to another location in `tools/` though i think

components/shared/code/BUCK
joshlafleur-zip: Glob being relative already makes it a generic rule. Should be moved and clarified granted

components/shared/code/BUCK
joshlafleur-zip: Yes to export the files out to make access easier. The users just select which file exported here

components/bms_boss/BUCK
joshlafleur-zip: Ahh wait a sec I didnt push my latest, I got somewhere a bit less ret

components/bms_boss/BUCK
joshlafleur-zip: LMFAO yea dude I know, although tbh its only these 3 files needed to preprocess the linkscript

components/shared/code/defs.bzl
rickyelopez: yeah but you're limiting the functionality here, right? This function is only capable of globbing/remapping ALL header files in the given dirs. What if I just want a subset? If you pass in the list of files as a parameter you get to be more precise if you want to.

components/shared/code/BUCK
rickyelopez: yeah I wasn't commenting on whether the glob was relative or not just the utility of it

components/shared/code/BUCK
rickyelopez: but these all get compiled, right? so you're potentially compiling files the user won't actually care about. They won't end up in the final binary because the compiler and linker can both drop them, but you're compiling them for nothing

components/bms_boss/BUCK
rickyelopez: well, it's only 3 files today lol, who knows what kind of sketchyness we will pull in the future

PR #335, opened 2025-09-14
vcrear: Listen to STW direction command

JoshLafleur:
### Describe changes

1. Add ability for STW to command VCREAR in reverse

### Impact

1. Driver can transition into reverse and back into drive when FEATURE_REVERSE is enabled

### Test Plan

- [x] Flash vcrear and vcfront
- [x] Enter TS On, ensure no torque can be commanded
- [x] Enter run mode, ensure car goes forward
- [x] Enter reverse, ensure car goes backwards
- [x] Enter forward, ensure car goes forward
- [x] Cycle through each direction again confirming the direction is as intended
- [x] Ensure the screen displays the correct direction
- [x] Exit TS On, re-enter run mode in reverse, ensure car goes in reverse
- [x] Enter forward, ensure car goes forward
- [x] Ensure vcrear commands maximum torque 15Nm when WOT
- [x] Ensure vcrear commands 15Nm max in reverse in run mode
- [x] Ensure vcfront maps pedal input to 0-15Nm when in reverse regardless of race mode

### DIFF
--- a/components/vc/front/src/torque.c
+++ b/components/vc/front/src/torque.c
@@ -23,6 +23,7 @@
  ******************************************************************************/
 
 #define DEFAULT_TORQUE_PITS 25.0f
+#define DEFAULT_TORQUE_LIMIT_REVERSE 15.0f
 
 #define ABSOLUTE_MAX_TORQUE 130.0f
 #define ABSOLUTE_MIN_TORQUE 0.0f
@@ -119,6 +120,15 @@ static void torque_periodic_100Hz(void)
             torque_data.torque_request_max = DEFAULT_TORQUE_PITS;
         }
     }
+
+#if FEATURE_IS_ENABLED(FEATURE_REVERSE)
+    CAN_gear_E direction = CAN_GEAR_FORWARD;
+    const bool direction_valid = CANRX_get_signal(VEH, STW_gear, &direction) == CANRX_MESSAGE_VALID;
+    if (direction_valid && (direction == CAN_GEAR_REVERSE))
+    {
+        torque_data.torque_request_max = DEFAULT_TORQUE_LIMIT_REVERSE;
+    }
+#endif
 
     torque = (bppc_getState() == BPPC_OK) ?
               apps_getPedalPosition() * torque_data.torque_request_max :
--- a/network/definition/data/components/vcfront/vcfront-rx.yaml
+++ b/network/definition/data/components/vcfront/vcfront-rx.yaml
@@ -2,14 +2,12 @@ messages:
   UDSCLIENT_vcfrontUdsRequest:
     sourceBuses: veh
     unrecorded: true
+  STW_driverControls:
+    sourceBuses: veh
 signals:
   VCPDU_vehicleState:
     sourceBuses: veh
   BMSB_bmsStatusMem:
     sourceBuses: veh
   BMSB_imdStatusMem:
-    sourceBuses: veh
-  STW_maxRequestTorque:
-    sourceBuses: veh
-  STW_raceMode:
     sourceBuses: veh
--- a/components/vc/front/BUCK
+++ b/components/vc/front/BUCK
@@ -126,6 +126,8 @@ cxx_library(
             "APP_V1_FeatureSels.yaml": "//components/shared:FeatureSels/APP_V1_FeatureSels.yaml",
             "Application_FeatureDefs.yaml": "//components/shared:FeatureDefs/Application_FeatureDefs.yaml",
             "SharedApp_FeatureDefs.yaml": "//components/shared/code:SharedApp_FeatureDefs.yaml",
+            "VC_FeatureSels.yaml": "//components/vc/shared/features:VC_FeatureSels.yaml",
+            "VC_FeatureDefs.yaml": "//components/vc/shared/features:VC_FeatureDefs.yaml",
             "FeatureDefs.yaml": "FeatureDefs.yaml",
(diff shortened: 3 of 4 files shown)


components/vc/rear/src/mcManager.c
```diff
@@ -101,10 +104,22 @@ static void mcManager_periodic_100Hz(void)
...
+                CAN_gear_E direction = CAN_GEAR_FORWARD;
 
                 const bool command_valid = CANRX_get_signal(VEH, VCFRONT_torqueRequest, &torque_command) == CANRX_MESSAGE_VALID;
+                const bool direction_valid = CANRX_get_signal(VEH, STW_gear, &direction) == CANRX_MESSAGE_VALID;
                 (void)CANRX_get_signal(VEH, VCFRONT_torqueManagerState, &manager_state);
 
+                if (direction_valid && (direction == CAN_GEAR_REVERSE))
+                {
+                    mcManager_data.direction = MCMANAGER_REVERSE;
+                    torque_command = SATURATE(0.0f, torque_command, MCMANAGER_TORQUE_LIMIT_REVERSE);
```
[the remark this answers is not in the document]
rickyelopez: sadge

components/vc/front/src/torque.c
```diff
@@ -120,6 +121,15 @@ static void torque_periodic_100Hz(void)
         }
     }
 
+#if FEATURE_IS_ENABLED(FEATURE_REVERSE)
```
rickyelopez, line 124: god I hate that you have to use `FEATURE_IS_ENABLED`... I need to look through the code for this feature tree stuff. If you just define these as `#define FTR_REVERSE true` in your codegen, you can then just do `#if FTR_REVERSE`

components/vc/front/src/torque.c
```diff
@@ -120,6 +121,15 @@ static void torque_periodic_100Hz(void)
...
+    CAN_gear_E direction = CAN_GEAR_FORWARD;
+    const bool direction_valid = CANRX_get_signal(VEH, STW_gear, &direction) == CANRX_MESSAGE_VALID;
+    if (direction_valid && (direction == CAN_GEAR_REVERSE))
+    {
+        torque_data.torque_request_max = DEFAULT_TORQUE_LIMIT_REVERSE;
+    }
```
rickyelopez, line 130: kinda meta and we can do this later, but what if instead of doing it this way we actually develop the concept of "gears"? Like, the vehicle starts in `GEAR_NEUTRAL` and then the steering wheel can submit "gear change requests" which the VC can then decide whether to honor or not.

That way, you can introduce additional checks in the VC and everything flows nicely.

So it would be like `STW_gearShiftRequest` which could be `GEAR_DRIVE`, `GEAR_NEUTRAL` or 	`GEAR_REVERSE`. Receive that in the VC, add whatever other checks you want (I can't think of any examples right now but I'm sure there are some) and then do the shift if the conditions are all good. Then you publish like a `VCREAR_gear` or something, and you can display that on the screen

rickyelopez (approved):
Fine in general but left some high level comments as to whether this is really the way you want to do this. Let's chat if you don't immediately understand what I'm saying here

components/vc/front/src/torque.c
JoshLafleur: _Technically_ the feature is defined as 1 so it will correctly evaluate. Thought itd be better to be verbose and allow `FEATURE_IS_ENABLED(...)` and `FEATURE_IS_DISABLED(...)`

JoshLafleur: I originally kept it seperate to distinguish from ENABLED features, and "FEATURES" that are enums (see APP_COMPONENT_ID iirc)

network/NetworkGen/templates/message_unpack.mako
```diff
@@ -90,7 +90,7 @@
         %if dtype == 64:
     int64_t tmp_${signal.name} = (m->u64 >> ${signal.start_bit}U) & ${(2**signal.native_representation.bit_width) - 1}U;
         %else:
-    int${math.ceil(signal.native_representation.bit_width / 8) * 8}_t tmp_${signal.name} = (m->u${dtype}[${idx_s}U] >> ${signal.start_bit % dtype}U) & ${(2**signal.native_representation.bit_width) - 1}U;
+            int${math.ceil(signal.native_representation.bit_width / 8) * 8}_t tmp_${signal.name} = (int${math.ceil(signal.native_representation.bit_width / 8) * 8}_t)((m->u${dtype}[${idx_s}U] >> ${signal.start_bit % dtype}U) & ${(2**signal.native_representation.bit_width) - 1}U);
```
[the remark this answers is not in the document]
JoshLafleur: :dang:

network/definition/data/components/vcfront/vcfront-rx.yaml
```diff
@@ -2,14 +2,12 @@ messages:
   UDSCLIENT_vcfrontUdsRequest:
     sourceBuses: veh
     unrecorded: true
+  STW_driverControls:
```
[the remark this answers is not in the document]
JoshLafleur: The distinction comes at the yamcan layer. Receiving a message will receive all signals and store them into memory. Receiving a single signal will only populate that into memory. Its a helper for cases where multiple signals/an entire message is to be received/bridging messages

components/vc/front/src/torque.c
JoshLafleur: Lets chat and revisit in a follow up ticket. Might be a good thing to get a newer member on so they can get their feet wetter

PR #337, opened 2025-09-15
vcrear: Add minimum torque when car is in motion

JoshLafleur:
### Describe changes

Currently there is significant torque shudder when no load conditions are present and the driver is not requesting torque. It is suspected that slack in the chain is causing control issues within the motor controller.

### Impact

1. When in motion and torque is requested, add a minimumtorque request such that the car does not accelerate but keeps the chain loaded.

### Test Plan

- [x] Enter run mode, push the car and ensure no torque is requested
- [x] Enter run mode, accelerate the car, ensure a minimum torque is requested when off the gas
- [x] When minimum torque is requested, ensure there is no acceleration
- [x] Ensure once the driver breaks and comes off the brake, the minimum torque is cut
- [x] Accelerate and re-enable minimum torque and hit the ESTOP. Re-enter Run Mode and ensure minimum torque is disabled

### DIFF
--- a/components/vc/rear/src/mcManager.c
+++ b/components/vc/rear/src/mcManager.c
@@ -21,6 +21,10 @@
 #define MCMANAGER_TORQUE_LIMIT 130.0f
 #define MCMANAGER_TORQUE_LIMIT_REVERSE 15.0f
 
+#define LASH_TORQUE 2.0f
+#define LASH_TORQUE_RPM_DISABLE 180.0f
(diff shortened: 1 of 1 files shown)


JoshLafleur:
I think if this doesnt even improve the shudder, its a worthwhile improvement to make because of the reduction in chain looseness -> Prevents the jerky-ness when the motor is unloaded

PR #344, opened 2025-09-19
Buck2: fix gcc version selection

rickyelopez:
### Describe changes

Group gcc releases by host arch/os and separate abi selection into its own field

### Impact

buck2 should select the correct release for the given host and target

### Test Plan


JoshLafleur (approved):
(╯°□°)╯︵ ┻━┻

PR #347, opened 2025-09-19
github: Run workflows on macos runners for firmware

JoshLafleur:
### Describe changes

Build buck2 firmware on macos runners

### Test Plan

- [ ] Did it pass big dawg?

### DIFF
--- a/.github/workflows/standard_checks.yml
+++ b/.github/workflows/standard_checks.yml
@@ -45,9 +45,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner: [ ubuntu-24.04, ubuntu-24.04-arm,]
+        runner: [ ubuntu-24.04, ubuntu-24.04-arm,
+                  macos-26, macos-15, macos-13 ]
                   # TODO: Bringup support
-                  # macos-15, macos-14, macos-13,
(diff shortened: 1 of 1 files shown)


.github/workflows/standard_checks.yml
```diff
@@ -45,9 +45,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runner: [ ubuntu-24.04, ubuntu-24.04-arm,]
+        runner: [ ubuntu-24.04, ubuntu-24.04-arm,
+                  macos-26, macos-15, macos-14, macos-13 ]
```
rickyelopez, line 49:
```suggestion
        runner:
          - ubuntu-24.04
          - ubuntu-24.04-arm
          - macos-26
          - macos-15
          - macos-14
          - macos-13
```

also, do we really need to support so many versions of macos? I'd say just do 26 and 15 unless you know of someone that is using one of the older ones

rickyelopez (approved):

.github/workflows/standard_checks.yml
JoshLafleur: dropped 14, but kept 15+26 since being different arm64 arch's and 13 being an x86 mac runner

PR #350, opened 2025-09-20
buck2: fix stw build

rickyelopez:
### Describe changes

fix stw build

### Impact

stw build not broken

### Test Plan

- [ ] stw builds


JoshLafleur (approved):
We should probably have this in github builds to make sure we dont break it by accident

rickyelopez:
> We should probably have this in github builds to make sure we dont break it by accident

yeah, I think I'm gonna pick up some work on it, I'll add it to ci if I do so

PR #358, opened 2025-09-25
Jl/smart date

JoshLafleur:
### Describe changes

Allows verifying CRC and possibly skipping a download if the binary is upto date, allows batch updating ECUs (currently a stupid way all single threaded), and allows specifying fast downloads

### Impact

Improved updatability

### Test Plan

- [x] Run a batch update on all ECUs, ensure no ECU is skipped
- [x] Rerun a batch update, ensure all ECUs are skipped


drive-stack/conUDS/src/modules/uds.rs
```diff
@@ -226,34 +227,54 @@ impl UdsClient {
...
         ];
 
-        info!("Sending ECU read DID command: {:02x?}", buf);
-
         match CanioCmd::send_recv(&buf, self.uds_queue_tx.clone(), 50)
-            .await?
+            .await
+            .map_err(|e| {
+                error!("send_recv failed: {e}");
+                UdsError::GeneralReject
```
rickyelopez, line 234: `GeneralReject` is a very specific UDS response, it shouldn't just be used for a general failure here unless the UDS response was actually `GeneralReject`. Same below

drive-stack/conUDS/src/arguments.rs
```diff
@@ -13,7 +13,11 @@ use crate::SupportedResetTypes;
 pub struct Arguments {
     /// the node to interact with
     #[argh(option, short = 'n')]
-    pub node: String,
+    pub node: Option<String>,
+
+    /// the node to interact with
```
rickyelopez, line 18: comment seems wrong

drive-stack/conUDS/src/arguments.rs
```diff
@@ -35,17 +39,31 @@ pub enum ArgSubCommands {
...
+    Batch(SubArgBatch),
 }
 
 /// Download an application to an ECU
 #[derive(Debug, FromArgs)]
 #[argh(subcommand, name = "download")]
 pub struct SubArgDownload {
+    /// only perform download if there is a CRC mismatch
+    #[argh(switch, short = 'c')]
+    pub verify_crc: bool,
```
rickyelopez, line 51: call this `force` and default to true

drive-stack/conUDS/src/main.rs
```diff
@@ -46,68 +50,220 @@ async fn main() {
...
 
-    let uds_node = cfg.nodes.get(&args.node).unwrap_or_else(|| {
-        error!("UDS node not defined");
-        std::process::exit(1)
-    });
-    debug!("UDS node identified: {:#?}", uds_node);
+    match args.subcmd {
+        ArgSubCommands::Batch(batch) => {
+            if batch.updates.is_empty() {
+                error!("No updates provided. Use -u node:/path/to/bin (repeatable).");
```
rickyelopez, line 56: I'd call this targets instead of updates I think

drive-stack/conUDS/src/main.rs
```diff
@@ -46,68 +50,220 @@ async fn main() {
...
+                        // proceed with download anyway
+                    }
+                }
+
+                uds.client.start_persistent_tp().await;
+                let _ = uds.reset_node(SupportedResetTypes::Hard).await;
+
+                if let Err(e) = uds.file_download(&bin, 0x08002000).await {
+                    error!("Download failed for node '{}': {}", node, e);
+                }
```
rickyelopez, line 136:
```suggestion
                } else {
				    info!("Download successful...")
				}
```

drive-stack/conUDS/src/main.rs
```diff
@@ -46,68 +50,220 @@ async fn main() {
...
+
+            if dl.verify_crc {
+                let mut file = File::open(dl.binary.to_path_buf()).expect("Binary does not exist!");
+                file.seek(SeekFrom::End(-4)).expect("Seek failed");
+                let mut buffer = [0u8; 4];
+                file.read_exact(&mut buffer).expect("CRC read failed");
+                let app_crc = u32::from_le_bytes(buffer);
+                println!("Application CRC to download: 0x{:08X}", app_crc);
+
+                let resp = uds.client.did_read(0x0003).await;
```
rickyelopez, line 182: same here re: 0x0003

drive-stack/conUDS/src/main.rs
```diff
@@ -46,68 +50,220 @@ async fn main() {
...
+            });
+            let uds_node = cfg.nodes.get(&node).unwrap_or_else(|| {
+                error!("UDS node '{}' not defined", node);
+                std::process::exit(1)
+            });
+
+            let mut uds = UdsSession::new(&args.device, uds_node.request_id, uds_node.response_id, args.fast).await;
+            info!("Downloading binary {:?} to node '{}'", dl.binary, node);
+
+            if dl.verify_crc {
```
rickyelopez, line 174: this seems like mostly copy-pasta from batch, maybe write a function for this?

drive-stack/conUDS/src/main.rs
```diff
@@ -46,68 +50,220 @@ async fn main() {
...
+                let mut file = File::open(dl.binary.to_path_buf()).expect("Binary does not exist!");
+                file.seek(SeekFrom::End(-4)).expect("Seek failed");
+                let mut buffer = [0u8; 4];
+                file.read_exact(&mut buffer).expect("CRC read failed");
+                let app_crc = u32::from_le_bytes(buffer);
```
rickyelopez, line 179: technically you should read the crc location from the app header at the start of the file instead of assuming it will be in the last four bytes. :footgun:

JoshLafleur: Yea, this is a bit tricky for the bootloader verification case, but im technically not even supporting that presently. At this moment (and as I understand commonly) validation in added to the end of payloads so making this assumption here _seems_ fine

drive-stack/conUDS/src/main.rs
JoshLafleur: Yes need a bit of clean up here

drive-stack/conUDS/src/arguments.rs
JoshLafleur: cant default to true, so changed to no skip

JoshLafleur:
Split into a crc verification/batch flashing PR (this) and a followup to bring up batch flashing in buck2

PR #364, opened 2025-09-28
buck2: Bringup openocd support

JoshLafleur:
### Describe changes

Adds support for debugging in buck2

### Test Plan

- [x] Build and debug a controller through an stlink
- [x] Execute `b main`, `monitor reset halt`, and then `c` and ensure it halts on the start of the `main` function


toolchains/gcc/defs.bzl
```diff
@@ -60,6 +60,12 @@ GccDistributionInfo = provider(
     },
 )
 
+GdbToolInfo = provider(
+    fields = {
+        "gdb": provider_field(typing.Any),
+    },
+)
```
rickyelopez, line 67: I don't think you need to do this. Instead, just add gdb to the `binary_utilities_info` on line 288 and then have your `toolchain` on line 155 of `openocd/defs.bzl` take a `CxxToolchainInfo`. I think that should work?

rickyelopez (dismissed):

Louiscars190:
image

1. Built VCPDU using Buck2
2. Executed `buck2 run components/vc/pdu:gdb-0` while STLink is connected to board
3. Added breakpoint main, ran monitor reset halt, and continued

PR #371, opened 2025-10-01
Jl/nvm task

JoshLafleur:
### Describe changes

1. Describe changes...

### Impact

1. Describe impact...

### Test Plan


components/shared/code/libs/lib_nvm.c
```diff
@@ -268,7 +306,24 @@ bool lib_nvm_nvmInitializeNewBlock(void)
 
 void lib_nvm_requestWrite(lib_nvm_entryId_E entryId)
 {
-    records[entryId].writeRequired = true;;
+    records[entryId].writeRequired = true;
+
+#if FEATURE_IS_ENABLED(NVM_TASK)
+    if (data.rtosStarted)
```
JoshLafleur, line 312: check if already requested

components/shared/code/HW/HW_flash.c
```diff
@@ -116,15 +120,23 @@ bool FLASH_erasePages(uint32_t startPageAddr, uint16_t pages)
...
         SET_REG(FLASH_AR, currPageAddr);
         // start the erase process
         SET_REG(FLASH_CR, GET_REG(FLASH_CR) | FLASH_CR_START);
 
         while (FLASH_BUSY())
-        {}
+        {
+#if FEATURE_IS_ENABLED(HW_FLASH_DELAY)
+            vTaskDelay(pdMS_TO_TICKS(1));
+#endif
```
JoshLafleur, line 138: CHeck if rtos is running

components/shared/code/libs/lib_nvm.c
```diff
@@ -227,30 +242,65 @@ void lib_nvm_init(void)
...
+                                           sizeof(lib_nvm_entryAction_S),
+                                           data.entry_action_queue_buff,
+                                           &data.entry_action_queue);
+#endif
+}
+
+void lib_nvm_start(void)
+{
+#if FEATURE_IS_ENABLED(NVM_TASK)
+    data.rtosStarted = true;
```
rickyelopez, line 257: I don't understand why this is necessary. None of these tasks will run until the rtos is started, right?

components/shared/code/libs/lib_nvm.c
```diff
@@ -227,30 +242,65 @@ void lib_nvm_init(void)
...
+void lib_nvm_dequeue(void)
+{
+    for (;;)
     {
-        recordWrite(NVM_ENTRYID_LOG);
+        lib_nvm_entryAction_S action;
+
+        if(xQueueReceive(data.queue_handle,
+                         &action,
+                         portMAX_DELAY))
```
rickyelopez, line 282: shouldn't the delay be 0?

components/shared/code/libs/lib_nvm.c
```diff
@@ -227,30 +242,65 @@ void lib_nvm_init(void)
...
+        {
+            if ((action.entryId == NVM_ENTRYID_LOG) &&
+                uxQueueMessagesWaiting(data.queue_handle))
+            {
+                xQueueSend(data.queue_handle,
+                           (void*)&action,
+                           portMAX_DELAY);
+            }
```
rickyelopez, line 290: what are you doing here? I don't get it. By this point you already know that there's an action queued and you've extracted it into `action` above. `uxQueueMessagesWaiting` returns the number of entries in the queue, so you're just checking if there's still messages left here? And then re-queing the action? What am I missing?

components/shared/code/libs/lib_nvm.c
```diff
@@ -268,7 +318,29 @@ bool lib_nvm_nvmInitializeNewBlock(void)
 
 void lib_nvm_requestWrite(lib_nvm_entryId_E entryId)
 {
-    records[entryId].writeRequired = true;;
+    if (records[entryId].writeRequired)
+    {
+        return;
+    }
```
rickyelopez, line 324: why?

components/shared/code/libs/lib_nvm.c
```diff
@@ -268,7 +318,29 @@ bool lib_nvm_nvmInitializeNewBlock(void)
...
+    if (data.rtosStarted)
+    {
+        lib_nvm_entryAction_S action = {
+            .entryId = entryId,
+            .action = NVM_ACTION_WRITE,
+        };
+        xQueueSend(data.queue_handle,
+                   (void*)&action,
+                   portMAX_DELAY);
+    }
```
rickyelopez, line 338: I'm lost

components/shared/code/libs/lib_nvm.c
```diff
@@ -308,19 +380,27 @@ uint32_t lib_nvm_getTotalCycles(void)
 
 void lib_nvm_cleanUp(void)
 {
-    portENTER_CRITICAL();
     data.mcuShuttingDown = true;
+    lib_nvm_check();
+    while(lib_nvm_writesRequired());
```
rickyelopez, line 385: Won't this block and prevent the writes from happening?

components/shared/code/libs/lib_nvm.c
```diff
@@ -347,11 +427,11 @@ static void initializeEmptyRecords(void)
     }
 }
 
-static void evaluateWriteRequired(lib_nvm_entry_t entryId)
+static bool evaluateWriteRequired(lib_nvm_entry_t entryId)
 {
     if (records[entryId].writeRequired)
     {
-        return;
+        return records[entryId].writeRequired;
```
rickyelopez, line 434:
```suggestion
        return true;
```

components/shared/code/libs/lib_nvm.c
```diff
@@ -365,6 +445,7 @@ static void evaluateWriteRequired(lib_nvm_entry_t entryId)
         records[entryId].writeRequired = true;
     }
```
rickyelopez, line 446: I'm not in love with this idea of checking if writes are required from within the nvm lib... Shouldn't the application using the lib be the one to determine if a write is required? _maybe_ there's a case for double-checking in the lib that a write hasn't been requested with exactly the same data, but otherwise this seems fishy to me

components/shared/code/libs/lib_nvm.c
```diff
@@ -378,39 +459,53 @@ static void recordWrite(const lib_nvm_entry_t entryId)
...
-    lib_nvm_blockHeader_S * const blockHeaderCurrent = (lib_nvm_blockHeader_S * const)getBlockBaseAddress((uint32_t)data.currentPtr);
     uint16_t entrySize = lib_nvm_entries[entryId].entrySize;
 
-    if (blockHeaderCurrent != (lib_nvm_blockHeader_S * const)getBlockBaseAddress((uint32_t)data.currentPtr + sizeof(lib_nvm_recordHeader_S) + entrySize))
+#if FEATURE_IS_ENABLED(NVM_TASK)
+    if (data.rtosStarted)
+    {
+        xSemaphoreTake(data.flash_mutex, portMAX_DELAY);
+    }
+#endif
```
rickyelopez, line 470: yeah this just seems like a critical section to me. I don't quite get why we need a semaphore here

components/shared/code/libs/lib_nvm.c
```diff
@@ -378,39 +459,53 @@ static void recordWrite(const lib_nvm_entry_t entryId)
...
+    uint32_t current_hdr = (uint32_t)data.currentPtr;
+    uint32_t current_record = (uint32_t)((lib_nvm_recordHeader_S*)data.currentPtr + 1);
+    lib_nvm_blockHeader_S * const block_header = (lib_nvm_blockHeader_S * const)getBlockBaseAddress(current_hdr);
+    data.currentPtr += sizeof(lib_nvm_recordHeader_S) / sizeof(storage_t);
+    data.currentPtr += lib_nvm_entries[entryId].entrySize / sizeof(storage_t);
+#if FEATURE_IS_ENABLED(NVM_TASK)
+    if (data.rtosStarted)
+    {
+        xSemaphoreGive(data.flash_mutex);
```
rickyelopez, line 479: wait, the semaphore isn't protecting the actual write?

components/shared/code/libs/lib_nvm.c
rickyelopez, line 1: I didn't get a chance to review your actual nvm PR, but am I correctly understanding that NVM is effectively non-contiguous in this design? I.e. you've separated the data you're writing into nvm into different records, and each record can be at basically a random location in the nvm page? I'm not in love with this. Seems like it would make it harder to check everything on startup, harder to invalidate data, harder to migrate pages, etc.

components/shared/code/libs/lib_nvm.c
JoshLafleur: Always delay nvm record log writes until it is the last thing to write. It basically ensures we dont write it, write something else, and then have to write it again because the number of writes increased

components/shared/code/libs/lib_nvm.c
JoshLafleur: Yes, which is what you want on cleanup. A calling task shouldnt continue until the nvm writes have all cleaned up, this is used by uds prior to resets

components/shared/code/libs/lib_nvm.c
JoshLafleur: We could only run this on things like the record and cycle log.

Did it this way so that the application can have an "assurance" that when it updates ram, it will seamlessly save data in the worse case of a system reset. For set and forget values, the application can request a write when its done transforming the data and then be assured that at the next time its needed, it will be ready for use

components/shared/code/libs/lib_nvm.c
JoshLafleur: Yea i think with how it eveolved thats maybe valid, unclear havent danced around it too much mentally

components/shared/code/libs/lib_nvm.c
rickyelopez: So is this actually just an indication that NVM setup is done? If so, maybe update the name to reflect that and make it accessible from other modules either with a global or a accessor.

`nvmInitComplete` or something

components/shared/code/libs/lib_nvm.c
rickyelopez: I see, makes sense. You need to document all of this a lot better I think

PR #372, opened 2025-10-01
Changed signed/unsigned logic

LucasZahlan:
### Describe changes

1. Changed signed/unsigned logic to check for manual assignment and takes the range into account for default assignment.
2. Removed mutable default argument
3. Added a compiler check to prevent manual assignment of unsigned assignment when a defined range is less than 0

### Impact

1. Improves default assignment accuracy based on range while still defaulting unsigned if it is able to be.
2. Remove uncertainty and possibility for issues with mutable default argument.
3. Allow for quicker troubleshooting if a user accidentally implements a negative unsigned message manually.

### Test Plan

- [ ] Attempt to drive a negative unsigned message.
- [ ] Check default behavior of messages with positive and negative ranges.
- [ ] Ensure manually assigned messages are not overridden by defaults

### DIFF
--- a/network/NetworkGen/classes/Can.py
+++ b/network/NetworkGen/classes/Can.py
@@ -76,7 +76,8 @@ class NativeRepresentation:
     Takes in a dictionary parsed from YAML
     """
 
-    def __init__(self, signal_def=dict()):
+    def __init__(self, signal_def=None):
+        signal_def = signal_def or {}
         NATREP_SCHEMA.validate(signal_def)
         if "endianness" in signal_def:
             try:
@@ -94,12 +95,25 @@ class NativeRepresentation:
         if "signedness" in signal_def:
             self.signedness = Signedness[signal_def["signedness"]]
         else:
-            self.signedness = Signedness.unsigned
+            self.signedness = (
+                Signedness.signed
+                if (self.range is not None and getattr(self.range, "min", 0) < 0)
+                else Signedness.unsigned
+            )
+
         if "endianness" in signal_def:
             self.endianness = Endianess[signal_def["endianness"]]
         else:
             self.endianness = Endianess.little
         self.resolution = get_if_exists(signal_def, "resolution", float, None)
+        if (
+            self.signedness == Signedness.unsigned
+            and self.range is not None
+            and getattr(self.range, "min", 0) < 0
+        ):
+            raise Exception(
+                f"Native representation cannot be unsigned when the range minimum is negative (min={self.range.min})"
+            )
 
         if self.bit_width is not None and (self.bit_width < 1 or self.bit_width > 64):
             raise Exception(f"Native representation bit width must be between 1 and 64 bits")


network/NetworkGen/classes/Can.py
```diff
@@ -94,12 +95,25 @@ def __init__(self, signal_def=dict()):
...
             self.endianness = Endianess.little
         self.resolution = get_if_exists(signal_def, "resolution", float, None)
+        if (
+            self.signedness == Signedness.unsigned
+            and self.range is not None
+            and getattr(self.range, "min", 0) < 0
+        ):
+            raise Exception(
+                f"Native representation cannot be unsigned when the range minimum is negative (min={self.range.min})"
+            )
```
JoshLafleur, line 116: see above

JoshLafleur (commented):
Really we should be erroring [here](https://github.com/concordia-fsae/firmware/blob/0ed29b0079ff6b9a3618c941a00cca1a9dca62aa/network/NetworkGen/classes/Can.py#L345) if this ever happens, we need to fix the native representation parsing [here](https://github.com/concordia-fsae/firmware/blob/0ed29b0079ff6b9a3618c941a00cca1a9dca62aa/network/NetworkGen/classes/Can.py#L306) so that the offset/scale is always correct and unsigned.

We should also make sure that an appropriate offset and scale is set by [this](https://github.com/concordia-fsae/firmware/blob/0ed29b0079ff6b9a3618c941a00cca1a9dca62aa/network/NetworkGen/classes/Can.py#L292) point.

Looking at the codegen files (pack and unpack) [here](https://github.com/concordia-fsae/firmware/blob/0ed29b0079ff6b9a3618c941a00cca1a9dca62aa/network/NetworkGen/templates/message_unpack.mako#L80) as well and making sure these values are correct will also be good.

For validation you will want to output likely verify with BMSB current

tools/yamcan/templates/message_unpack.mako
```diff
@@ -101,12 +101,14 @@
...
+    sigrx->${sig_name} = (${signal.datatype.value})(
+        ((${signal.datatype.value})tmp_${signal.name}) * ${float(signal.scale)}f
+        + ${float(signal.offset)}f
+    );
     %elif "int" in signal.datatype.value:
     sigrx->${sig_name} = (tmp_${signal.name} * ${int(signal.scale)}) + (${int(signal.offset)});
     %endif
 %else:
-    (void)m;
-    (void)val;
```
JoshLafleur: Why did this diff? If youre in this case, you should need to disregard both function args, maybe im missing something?

tools/yamcan/templates/message_unpack.mako
```diff
@@ -101,12 +101,14 @@
     tmp_${signal.name} = (tmp_${signal.name} & ${(2**(signal.native_representation.bit_width - 1) - 1)}U) * (((tmp_${signal.name} & (1U << ${(signal.native_representation.bit_width - 1)})) != 0U) ? -1.0f : 1.0f);
     %endif
     %if "float" in signal.datatype.value:
-    sigrx->${sig_name} = (${signal.datatype.value})(tmp_${signal.name}) * ${float(signal.scale)}f + (${float(signal.offset)}f);
+    sigrx->${sig_name} = (${signal.datatype.value})(
+        ((${signal.datatype.value})tmp_${signal.name}) * ${float(signal.scale)}f
+        + ${float(signal.offset)}f
+    );
```
JoshLafleur: Maybe im unsure what this changes, but it seems like you added an internal cast to `tmp_${signal.name}` which was already being done it seems. Maybe you meant to correct outer brackets of the original cast instead of adding another cast?

tools/yamcan/templates/message_pack.mako
```diff
@@ -101,16 +101,39 @@ __attribute__((always_inline)) static inline void set_${bus.upper()}_${signal.me
...
+    )(
+        (((float)val) - ${float(signal.offset)}f) / ${float(signal.scale)}f
+    );
     %elif "int" in signal.datatype.value:
-    tmp_${signal.name} = (${'u' if signal.native_representation.signedness == Signedness.unsigned else ''}int${dtype}_t)((val - ${int(signal.offset)}) / ${int(signal.scale)});
+    tmp_${signal.name} = (
+        ${'u' if signal.native_representation.signedness == Signedness.unsigned else ''}int${dtype}_t
+    )(
+        (val - ${int(signal.offset)}) / ${int(signal.scale)}
+    );
```
JoshLafleur: This all looks the same except for a single float cast, but it doesnt seem like we need that?

tools/yamcan/templates/message_pack.mako
```diff
@@ -101,16 +101,39 @@ __attribute__((always_inline)) static inline void set_${bus.upper()}_${signal.me
 %>\
 %if signal.native_representation.bit_width == 1:
     atomicXorU8(&m->u8[${int(signal.start_bit / 8)}U], (uint${dtype}_t)((val ? 1U : 0U) << ${signal.start_bit % 8}U));
+    %elif signal.discrete_values:
+    %if signal.native_representation.signedness == Signedness.unsigned:
+    uint${dtype}_t tmp_${signal.name} = (uint${dtype}_t)val;
+    %else:
+    int${dtype}_t tmp_${signal.name} = (int${dtype}_t)val;
+    %endif
```
JoshLafleur: signed discrete values doesnt make sense. Discrete values are some integer value with some base offset

tools/yamcan/templates/message_pack.mako
```diff
@@ -101,16 +101,39 @@ __attribute__((always_inline)) static inline void set_${bus.upper()}_${signal.me
...
+    %if signal.native_representation.endianness.value == 0 and signal.native_representation.bit_width > 8:
+    reverse_bytes((uint8_t*)&tmp_${signal.name}, ${int(signal.native_representation.bit_width / 8)}U);
+    tmp_${signal.name} = ((tmp_${signal.name} & (${'u' if signal.native_representation.signedness == Signedness.unsigned else ''}int${dtype}_t)(~(255U))) >> ${(8 - signal.native_representation.bit_width % 8) % 8}U) | (int${dtype}_t)((tmp_${signal.name} & ${(2**(8 - (signal.native_representation.bit_width % 8))) - 1}U));
+    %endif
+    %if dtype == 64:
+    atomicXorU64(&m->u64, (uint64_t)(tmp_${signal.name} & ${(2**signal.native_representation.bit_width) - 1}U) << ${signal.start_bit}U);
+    %else: 
+    atomicXorU${dtype}(&m->u${dtype}[${idx_s}], (uint${dtype}_t)((uint${dtype}_t)(tmp_${signal.name}) & ${(2**signal.native_representation.bit_width) - 1}U) << ${signal.start_bit % dtype}U);
+    %endif
```
JoshLafleur: What are you exactly trying to accomplish here?

tools/yamcan/classes/Can.py
```diff
@@ -326,7 +400,7 @@ def calc_signal_params(self):
                     sig_range = nat_rep.range.max - nat_rep.range.min
                 elif nat_rep.signedness == Signedness.signed:
                     self.offset = 0
-                    sig_range = 2* max(nat_rep.range.max, nat_rep.range.min)
+                    sig_range = 2* max(nat_rep.range.max, -nat_rep.range.min)
```
JoshLafleur: This is only true if min is negative, although i think you did catch a bug here. nice

tools/yamcan/classes/Can.py
```diff
@@ -312,7 +312,81 @@ def calc_signal_params(self):
                 print(
                     f"Signal '{self.name}' uses discrete value table '{dv}' which has an invalid range"
                 )
-                self.is_valid = False
+            nat_rep.signedness = Signedness.unsigned
```
JoshLafleur: All of this and below applies to discrete values because of the previous check, but none of it actually is applicable to discrete values. DVs dont have concepts of floats etc, theyre enums.

To be fair, this is a vlid thing todo, but you shouldnt have most of the following logic

JoshLafleur (commented):
Please re-review the ticket and the associated pats PR comments

tools/yamcan/templates/message_unpack.mako
```diff
@@ -81,7 +81,7 @@
 %>\
 %if signal.native_representation.bit_width == 1:
     sigrx->${sig_name} = (m->u8[${int(signal.start_bit / 8)}U] & (1U << ${signal.start_bit % 8}U)) != 0U;
-%elif "float" in signal.datatype.value or "int" in signal.datatype.value: # Handles both int and uint
+%elif "float" in signal.datatype.value or "int" in signal.datatype.value:
```
JoshLafleur: Why was this diffed?

tools/yamcan/classes/Can.py
```diff
@@ -306,13 +306,60 @@ def calc_signal_params(self):
...
             if not nat_rep.range.is_valid:
-                print(
-                    f"Signal '{self.name}' uses discrete value table '{dv}' which has an invalid range"
+                raise Exception(
+                    f"Signal '{self.name}' uses discrete value table '{dv.name}' with invalid range "
+                    f"({dv.min_val}..{dv.max_val})"
                 )
-                self.is_valid = False
+
+            if not isinstance(nat_rep.bit_width, int) or not (1 <= nat_rep.bit_width <= 64):
```
JoshLafleur: I dont _think_ the typecheck here is required as the bit width should be restrained to an int in the schema

tools/yamcan/classes/Can.py
```diff
@@ -342,11 +389,11 @@ def calc_signal_params(self):
...
                     {"min": 0, "max": 2**nat_rep.bit_width * nat_rep.resolution}
                 )
+                self.scale = nat_rep.resolution
         else:
-            nat_rep = NativeRepresentation()
-            nat_rep.bit_width = 1
-            nat_rep.resolution = 1
-            nat_rep.range = Range({"min": 0, "max": 1})
+            raise Exception(
+                f"Signal '{self.name}' has neither discreteValues nor nativeRepresentation defined;"
```
JoshLafleur:
```suggestion
                f"Signal '{self.name}' has neither discreteValues nor nativeRepresentation defined"
```

tools/yamcan/templates/message_pack.mako
JoshLafleur: This was not addressed and shouldnt have been resolved

PR #378, opened 2025-10-06
scons: #fuckscons

JoshLafleur:
RIP pour one out for the homie

### Test Plan

- [ ] Does it still build?


rickyelopez (approved):
Damn, what a crazy turn of events

PR #381, opened 2025-10-07
bmsw & vcs(shared): Add support for task stack usage detection and transmission

KaileeSantucci:
### Describe changes

Added support for task stack usage into the BMSW and vcs (shared/RTOS)

### Test Plan

- [ ] Check the rtosTaskMemInfo ids ? 
- [ ]  Ensure the micro still runs


KaileeSantucci:
all VCs all tasks 200 bytes spare

PR #409, opened 2025-11-07
Adding support for 103JT Thermistors

### LINKED ISSUE #408: Bringup 103JT Thermistors for New TSB
[P9-JT-Thermistor.pdf](https://github.com/user-attachments/files/23424774/P9-JT-Thermistor.pdf)

Within `firmware/components/shared/code/libs/lib_thermistors.h`,
Create a new struct under `public vars` with the new B value, T0, R0 for the 103JT thermistors, used in CFR26 TSB, using data from linked datasheet.

Louiscars190:
### Describe changes

1. Added instance of `lib_thermistors_BParameter_S` in shared thermistor library to be used in new worker.

### Impact

1. Allows proper voltage to temperature conversion with 103JT thermistor's b parameter.

Resolves #408 
### Test Plan

- [ ] Once BMS Worker is assembled and validated, thermistors will be tested against ambient temperature of cells.
- [ ] Tested for all 6 workers (should be identical)
- [ ] Add config for worker to use `NTC103JT_bParam`

### DIFF
--- a/components/shared/code/libs/lib_thermistors.h
+++ b/components/shared/code/libs/lib_thermistors.h
@@ -45,6 +45,12 @@ static const lib_thermistors_BParameter_S NCP21_bParam = { // For the NCP21XV103
     .R0 = 10000U,
 };
 
+static const lib_thermistors_BParameter_S NTC103JT_bParam = { // For the 103JT-100 and 103JT-050
+    .B = 3435U,
+    .T0 = 25U + KELVIN_OFFSET,
+    .R0 = 10000U,
+};
+
 /******************************************************************************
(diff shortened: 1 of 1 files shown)


components/bms_worker/src/Environment.c
JoshLafleur: Remove this diff, youll be changing the workers we currently use. If you want to plan the work, create an issue for the BMS worker config 1 and add support for NTC103JT thermistor. This will need to be gated by feature tree

Louiscars190: Shit that's so facts I'm retarded

PR #414, opened 2025-11-16
yamcan: Cleanup folder structure

joshlafleur-zip:
### Describe changes

1. Moves yamcan from network folder, clarifiees definition name


network/BUCK
```diff
@@ -9,7 +9,7 @@ load(
 
 build_network(
     name = "network",
-    data_dir = "definition/",
+    data_dir = "vehicle-def/",
```
rickyelopez: idk if I agree with this... what was wrong with `definition`? I think the change that would actually be good here would be to support multiple different models. I.e. have like `definition/shared`, `definition/cfr25/...`, `definition/cfr26/...`, etc.

PR #439, opened 2025-12-23
Jl/launch control

JoshLafleur:
### Describe changes

Add launch control state machine

### Impact

1. Launch control can be entered through driver request. No impact to torque request presently

### Test Plan

- [x] Request launch control without brake pressed. Ensure the transition is rejected
- [x] Request launch control with brake pressed. Ensure HOLDING is entered
- [x] Release brake, ensure it exits to INACTIVE
- [x] Re-enter holding, press the throttle and ensure the BPPC doesnt fault
- [x] Ensure that we transition HOLDING->SETTLING->PRELOAD after 0.5s of the pedal held
- [x] Remove throttle (keeping the brake pressed) in PRELOAD, ensure we enter INACTIVE
- [x] Re-enter PRELOAD and release brake, ensure transition into LAUNCH
- [x] In LAUNCH, press brake, ensure it exits launch
- [x] In LAUNCH, release throttle, ensure it exits LAUNCH

### DIFF
--- a/components/vc/front/src/torque.c
+++ b/components/vc/front/src/torque.c
@@ -38,6 +38,7 @@
 #define LAUNCH_CONTROL_REJECTED_MS 5000
 
 #define LAUNCH_CONTROL_BRAKE_THRESHOLD 0.30
+#define LAUNCH_CONTROL_BRAKE_THRESHOLD_LAUNCH 0.05
 #define LAUNCH_CONTROL_THROTTLE_THRESHOLD 0.50
 #define LAUNCH_CONTROL_THROTTLE_START_CUTOFF 0.10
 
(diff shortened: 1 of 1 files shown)


components/vc/front/src/torque.c
```diff
@@ -47,13 +54,167 @@ static struct
...
+}
+
+static void evaluate_launch_control(float32_t accelerator_position, float32_t brake_position)
+{
+#if FEATURE_IS_ENABLED(FEATURE_LAUNCH_CONTROL)
+    CAN_digitalStatus_E launch_control_requested = CAN_DIGITALSTATUS_SNA;
+    const bool launch_control_request_was_active = torque_data.torque_control_request_active;
+    torque_data.torque_control_request_active = (CANRX_get_signal(VEH, SWS_requestLaunchControl, &launch_control_requested) != CANRX_MESSAGE_SNA) &&
+                                                (launch_control_requested == CAN_DIGITALSTATUS_ON);
+    const bool launch_control_request_rising = !launch_control_request_was_active && torque_data.torque_control_request_active;
```
rickyelopez: do you want to do this check all the time? Or just if you're in the INACTIVE state?

components/vc/front/src/torque.c
```diff
@@ -47,13 +54,167 @@ static struct
...
+                if ((brake_position > LAUNCH_CONTROL_BRAKE_THRESHOLD) &&
+                    (accelerator_position < LAUNCH_CONTROL_THROTTLE_START_CUTOFF))
+                {
+                    torque_data.launch_control_state = LAUNCHCONTROL_HOLDING;
+                    drv_timer_stop(&torque_data.launch_control_timer);
+                }
+                else
+                {
+                    torque_data.launch_control_state = LAUNCHCONTROL_REJECTED;
+                    drv_timer_start(&torque_data.launch_control_timer, LAUNCH_CONTROL_REJECTED_MS);
```
rickyelopez: maybe rename this timer? `launch_control_lockout_timer` maybe? Also I like to put the expected period in the timer name so you don't accidentally poll it from a function running at the wrong period (which would mess up the timer).

Also, aren't we camel case in this context? Shouldn't this be something like `launchControlLockoutTimer_10Hz`

components/vc/front/src/torque.c
```diff
@@ -47,13 +54,167 @@ static struct
...
+            {
+                torque_data.launch_control_state = LAUNCHCONTROL_INACTIVE;
+            }
+            break;
+        case LAUNCHCONTROL_PRELOAD:
+            if (accelerator_position < LAUNCH_CONTROL_THROTTLE_THRESHOLD)
+            {
+                torque_data.launch_control_state = LAUNCHCONTROL_INACTIVE;
+            }
+            else if (brake_position < LAUNCH_CONTROL_BRAKE_THRESHOLD)
```
rickyelopez: you should set a minimum threshold here too. Like make sure the driver releases the brakes below 5% or smth, not just to 29%. So do like a `LAUNCH_CONTROL_BRAKE_THRESHOLD_ENTER` and `LAUNCH_CONTROL_BRAKE_THRESHOLD_LAUNCH`

components/vc/front/src/torque.c
```diff
@@ -126,6 +287,13 @@ torque_launchControl_E torque_getLaunchControlState(void)
     return torque_data.launch_control_state;
 }
 
+bool torque_isLaunching(void)
+{
+    const bool launching = (torque_data.launch_control_state != LAUNCHCONTROL_INACTIVE) &&
+                           (torque_data.launch_control_state != LAUNCHCONTROL_REJECTED);
```
rickyelopez: same here, be explicit about allowed states in case you add a state in the future and forget to add it here.

components/vc/front/src/torque.c
```diff
@@ -228,75 +399,26 @@ static void torque_init(void)
...
-        }
-
-        torque_data.torque_request_max = torque_request_max;
-    }
-    else
-    {
-        drv_timer_stop(&torque_data.torque_change_timer);
-    }
-
+    float32_t torque_request_max = evaluate_torque_max();
```
rickyelopez: should this come before evaluating launch control?

JoshLafleur: It doesnt really matter here, there will be logic after max torque setting to take the launch control torque modulation, and then transform that relative to some traction algorithm

components/vc/front/src/torque.c
```diff
@@ -47,13 +54,167 @@ static struct
...
+            }
+            break;
+        case LAUNCHCONTROL_HOLDING:
+            if (brake_position < LAUNCH_CONTROL_BRAKE_THRESHOLD)
+            {
+                torque_data.launch_control_state = LAUNCHCONTROL_INACTIVE;
+            }
+            else if (accelerator_position > LAUNCH_CONTROL_THROTTLE_THRESHOLD)
+            {
+                drv_timer_start(&torque_data.launch_control_timer, LAUNCH_CONTROL_SETTLING_MS);
```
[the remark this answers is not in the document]
JoshLafleur: Couldnt we footgun with a singular timer here though? Like not transitioning one timer or another is the same same imo

components/vc/front/src/torque.c
JoshLafleur: Thats valid, so only enter launch once you have gotten under 5%?

PR #441, opened 2025-12-23
torque manager: Implement basic traction control pid to reduce torque…

JoshLafleur:
… request

### Describe changes

1. Describe changes...

### Impact

1. Describe impact...

### Test Plan


components/vc/front/src/torque.c
```diff
@@ -218,6 +231,53 @@ static void evaluate_launch_control(float32_t accelerator_position, float32_t br
...
+    epid_pi_sum(&torque_data.traction_control_pid, 0.0f, max_torque);
+    epid_util_ilim(&torque_data.traction_control_pid, 0.0f, TRACTION_CONTROL_ILIM);
+
+    return torque_data.traction_control_pid.y_out;
+}
+
+static void evaluate_traction_control(void)
+{
+    CAN_digitalStatus_E traction_control_requested = CAN_DIGITALSTATUS_SNA;
+    float32_t rear_axle_rpm = 0.0f;
```
rickyelopez: I could have sworn we were camel case. Did I miss a memo?

components/vc/front/src/torque.c
```diff
@@ -42,6 +43,14 @@
 #define LAUNCH_CONTROL_THROTTLE_THRESHOLD 0.50
 #define LAUNCH_CONTROL_THROTTLE_START_CUTOFF 0.10
 
+#define TRACTION_CONTROL_TARGET_SLIP 0.10f
+
+#define TRACTION_CONTROL_KP 800 // Starting Number: If we slip 1% extra, reduce 8Nm
+#define TRACTION_CONTROL_KI 1000   // Starting Number: If we integrate 1% slip error, reduce 10Nm
+#define TRACTION_CONTROL_ILIM (MIN_TORQUE_RANGE / TRACTION_CONTROL_KI)
+
+#define TRACTIONCONTROL_RAXLE_THRESHOLD_RPM 60
```
rickyelopez: Just use TC
```suggestion
#define TC_MIN_SPEED_THRESH 60 // [RPM] Minimum speed above which TC can activate
```

components/vc/front/src/torque.c
JoshLafleur: I think functions are camel, but variables are underscored? Atleast thats how I have always done it

JoshLafleur: Damn it must be nice being the one to dictate these things jeez louise. Shouldve came to school first IG

jkjk, I can get behind whatever as long as we formalize it, would you want to make a doc with these rules as oyu have them internalized, say in like `docs/guidelines.md` folder. I started something there a while ago and never followed through. Maybe get oyur input on it and add whetever, then once it set we can just continue forwards doing the right thing

components/vc/front/src/torque.c
```diff
@@ -61,7 +69,13 @@ static struct
     drv_timer_S torque_change_timer;
     drv_timer_S launch_control_timer;
 
-    torque_launchControl_E launch_control_state;
+    torque_launchControlState_E launch_control_state;
+    torque_tractionControlState_E traction_control_state;
```
JoshLafleur: camelCase

JoshLafleur:
Following this PR, there will be a singular baseline PID tuning to get a ball park tuning of this done in the loading dock. This will likely need to be further tuned, however this price will be OK to pay at a later point as there should be relatively correct orders of magnitude of the coefficients

PR #447, opened 2025-12-27
buck2: Pin prelude

JoshLafleur:
### Describe changes

1. Pin prelude to a working version

### Test Plan

- nah

### DIFF
--- a/.buckconfig
+++ b/.buckconfig
@@ -12,10 +12,13 @@
   fbcode_macros = none
   buck2 = none
 
-# Uses a copy of the prelude bundled with the buck2 binary. You can alternatively delete this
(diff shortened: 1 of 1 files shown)


JoshLafleur:
awww _but it builds on my machine :(_

JoshLafleur:
it seems to be something with https://github.com/dtolnay/install-buck2/blob/c3999b30bbeee68f02bea47034ecb07395d2fca2/action.yml#L30 always pulling the latest tag - in and of itself not stable. Might need to:
A. Open a PR with the action to support downloading specific tags from the action
B. Re-implement the action locally and support tagging it explicitly

PR #464, opened 2025-12-30
Improve PDU power manager and bringup PDU faults

JoshLafleur:
### Describe changes

1. Adds latcheable faults to the PDU power manager
2. Integrates fault manager into the PDU

### Test Plan

- [x] Ensure faults are output and update

### DIFF
--- a/components/vc/pdu/src/powerManager.c
+++ b/components/vc/pdu/src/powerManager.c
@@ -17,6 +17,7 @@
 #include "drv_inputAD.h"
 #include "drv_outputAD.h"
 #include "app_vehicleState.h"
+#include "app_faultManager.h"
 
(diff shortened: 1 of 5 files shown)


components/vc/pdu/src/powerManager.c
```diff
@@ -33,25 +60,79 @@ static struct
...
+
+    bool okBattery:1;
+    bool okLoads:1;
+    bool okSafety:1;
+
+    bool charged:1;
+    bool sleeping:1;
+    bool lowBattery:1;
+    bool overvoltage:1;
+} pm_data;
```
rickyelopez: just call it `pm` lol you're not adding any clarity here, just length

PR #465, opened 2025-12-30
Fix fault manager bit setting of faults waiting to transmit

JoshLafleur:
### Describe changes

1. Dumb roll


components/shared/code/app/app_faultManager.c
rickyelopez: I had built out the FLAGS interface which you could use here instead of re-building a subset

PR #481, opened 2026-01-01
Fix: Enforce brake and speed checks for Race Mode

### LINKED ISSUE #453: Add mode change requirements
Currently you can enter race mode while on the pedal and while moving. Change this to require the car to be stopped and the brake to be applied

luckys00:
## Summary
This PR implements safety checks for switching to Race Mode, as requested in issue #453.

## Changes
- Updated `evaluate_mode_change` in `torque.c` to accept `brake_position`.
- Added logic to ensure the vehicle speed is 0 AND the brake is pressed (>10%) before allowing the mode switch.
- Updated the function call in `torque_periodic_100Hz` to pass the brake position.

## Ticket
Fixes #453

### DIFF
--- a/components/vc/front/src/torque.c
+++ b/components/vc/front/src/torque.c
@@ -12,6 +12,7 @@
 #include "apps.h"
 #include "bppc.h"
 #include "Module.h"
+#include "app_vehicleSpeed.h"
 #include "ModuleDesc.h"
 #include "string.h"
 #include "lib_utility.h"
@@ -136,7 +137,7 @@ static bool evaluate_gear_change(float32_t accelerator_position, float32_t brake
     return ret;
 }
 
-static bool evaluate_mode_change(void)
+static bool evaluate_mode_change(float32_t brake_position)
 {
     bool ret = false;
     CAN_digitalStatus_E race_mode_change_requested = CAN_DIGITALSTATUS_SNA;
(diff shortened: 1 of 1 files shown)


components/vc/front/src/torque.c
```diff
@@ -134,7 +134,12 @@ static bool evaluate_mode_change(void)
                                           (race_mode_change_requested == CAN_DIGITALSTATUS_ON);
     const bool race_mode_change_rising = !race_mode_change_was_requested && torque_data.race_mode_change_active;
 
-    if (race_mode_change_rising)
+    const float32_t vehicleSpeed = wheelSpeed_getAxleRPM(AXLE_FRONT);
```
rickyelopez: feels like there should be a validity check here too, i.e. checking that the wheel speed from the front axle is valid

components/vc/front/src/torque.c
```diff
@@ -134,7 +134,12 @@ static bool evaluate_mode_change(void)
...
+    const bool safety_checks_ok = (brake_position > PEDAL_APPLIED_THRESHOLD) &&
+                                  (vehicleSpeed < VEHICLE_STOPPED_THRESHOLD);
+
+
+    if (race_mode_change_rising && safety_checks_ok)
```
rickyelopez: constants like this are free
```suggestion
    const bool brakePedalPressed = brake_position > PEDAL_APPLIED_THRESHOLD;
    const bool vehicleStationary = vehicleSpeed < VEHICLE_STOPPED_THRESHOLD;


    if (race_mode_change_rising && brakePedalPressed && vehicleStationary)
```

components/vc/front/src/torque.c
```diff
@@ -145,7 +146,12 @@ static bool evaluate_mode_change(void)
                                           (race_mode_change_requested == CAN_DIGITALSTATUS_ON);
     const bool race_mode_change_rising = !race_mode_change_was_requested && torque_data.race_mode_change_active;
 
-    if (race_mode_change_rising)
+    const float32_t vehicleSpeed = app_vehicleSpeed_getVehicleSpeed();
+
+    const bool brakePedalPressed = brake_position > PEDAL_APPLIED_THRESHOLD;
+    const bool vehicleStationary = vehicleSpeed < VEHICLE_STOPPED_THRESHOLD;
```
rickyelopez: probably use `>=` here

luckys00:
@rickyelopez 
Done! I updated the safety check to use >= and squashed everything into a single commit.

luckys00:
Hi @rickyelopez, just a gentle reminder to review this when you get a chance. I've addressed the feedback and squashed the commits. Thanks!

joshlafleur-zip (approved):
lgtm, but probably good to wait for ricky also (itll take a few days to get someone to test the code before merging).

Also would be good to make a test plan like https://github.com/concordia-fsae/firmware/pull/459, but tbh there I was a bit lazy.

Something like:
- With brakes applied at <10%, ensure reverse requests are rejected
- With the car moving over 0.5m/s, ensure reverse requests are rejected
- With the car stopped and with the brakes applied, ensure the vehicle transitions into reverse

@freeman803 could support testing, he will be at the workshop and I can reach out to him directly, lets validate the code before merging

PR #521, opened 2026-02-05
added messages to display on steering wheel

freeman803:
### Describe changes

1. Added messages to bmsb to help show max difference between cells

### Impact

1. allow display on steering wheel

### Test Plan

- [ ]Flash code and make sure messages are populated


network/definition/data/components/bmsb/bmsb-signals.yaml
```diff
@@ -350,3 +350,29 @@ signals:
...
       signedness: unsigned
     continuous: true
+
+  bmsMinVoltage:
+    unit: 'V'
+    description: min voltage in pack
+    nativeRepresentation:
+      endianness: little
+      bitWidth: 11
+      resolution: 0.01
```
JoshLafleur: An 11 bit can do 2000 count so your resolution or bitwidth is not efficient. Please correct for needed resolution and needed data size onto thebus

PR #526, opened 2026-02-26
move Units enum to units.yaml

Alr0611:
## What
Moves the `Units` enum out of `Types.py` into a `units.yaml` file under `network/definition/`...

## Changes
- `network/definition/units.yaml` — new file defining all valid units as key/symbol pairs
- `tools/yamcan/yamcan.py` — added `generate_units()` to load units from YAML at build time; replaced `Units` enum validation with a dict lookup against loaded units
- `tools/yamcan/classes/Types.py` — removed `Units` enum
- `tools/yamcan/classes/Can.py` — updated `CanSignal` to store unit as a plain string instead of a `Units` enum member
- `tools/yamcan/templates/dbc_functions.mako` — updated `signal.unit.value` → `signal.unit`
- `tools/yamcan/templates/message_unpack.mako` — updated `signal.unit.name` → `signal.unit`

## Testing
[x] Ran `python3 tools/yamcan/yamcan.py --build --data-dir network/definition` successfully
[x] Ran `--gen-dbc` and verified units appear correctly in generated DBC file
[x] Full `buck2 build :cfr25-embedded` passes


network/definition/units.yaml
```diff
@@ -0,0 +1,29 @@
...
+  Nm:      "Nm"
+  N:       "N"
+  deg:     "deg"
+  Hz:      "Hz"
+  Wb:      "Wb"
+  psi:     "PSI"
+  mps2:    "m/s2"
+  mps:     "m/s"
+  bytes:   "B"
+  Whr:     "Whr"
```
rickyelopez: is this watt-hour? if so: 
```suggestion
  Wh:     "Wh"
```

tools/yamcan/classes/Can.py
```diff
@@ -168,7 +168,7 @@ def __init__(
             signal_def, "sna", SnaParams, SnaParams(), extra_params=self.name
         )
         try:
-            self.unit = get_if_exists(signal_def, "unit", Units, Units.none)
+            self.unit = get_if_exists(signal_def, "unit", str, "")
```
rickyelopez: Is unit a mandatory field? If so, don't provide a default of `""` here

tools/yamcan/classes/Can.py
```diff
@@ -203,7 +203,7 @@ def crc32(self):
         crc = crc32(self.native_representation.signedness.value.encode(), crc)
         crc = crc32(self.native_representation.signedness.value.encode(), crc)
         crc = crc32(self.native_representation.endianness.value.to_bytes(4, 'little'), crc)
-        crc = crc32(self.unit.value.encode(), crc)
+        crc = crc32(self.unit.encode())
```
rickyelopez: this seems like a mistake

```suggestion
        crc = crc32(self.unit.encode(), crc)
```

tools/yamcan/yamcan.py
rickyelopez: you need to run black on your changed python files

tools/yamcan/yamcan.py
```diff
@@ -324,7 +330,8 @@ def process_node(node: CanNode):
                     raise Exception(f"Signal '{sig}' in '{node.name}' has neither a discreteValues or nativeRepresentation.")
                 if "unit" in definition:
                     try:
-                        Units(definition["unit"])
+                        if definition["unit"] not in valid_units:
+                            raise Exception(f"Unit '{definition['unit']}' is not an accepted unit.")
```
rickyelopez: Add a note about where to check for valid units here

```suggestion
                            raise Exception(f"Unit '{definition['unit']}' is not an accepted unit. See `path/to/units.yaml` for supported units.")
```

tools/yamcan/classes/Can.py
Alr0611: unit is optional several signals across the network have no unit (discrete/boolean) like stw-signals.yaml and sws-signals.yaml the `""` handles the cases where unit isnt specified

Alr0611: ur right for non discrete or non boolean values signals, a unit should always be specified (even if `none)` ill remove the `""` default and add validation to enforce unit is required for continuous signals and raise an error if a discrete or boolean has one set

network/definition/data/components/elcon/elcon-signals.yaml
```diff
@@ -30,3 +30,4 @@ signals:
         min: 0
         max: 256
     continuous: true  
+    unit: ''
```
rickyelopez: Imo, an explicit `none` here would be better

```suggestion
    unit: none
```

rickyelopez (commented):
Looks like there are a couple of my comments that still need to be addressed.

Separately, could you please squash your commits into logical groups and make sure that they are named descriptively?

tools/yamcan/classes/Can.py
Alr0611: i continued with `unit: ''` instead of `unit: none` because we validate against symbol values (e.g. `''`, `'V'`, `'%'`) not key names. Since signals across the repo already reference units by symbol, validating against values keeps everything consistent and avoids having to update every signal file. `unit: ''` maps to the `none` key in `units.yaml` which has value `""`

rickyelopez:
We use rebases in this repo to stay up to date with the main branch, not merges.

PR #549, opened 2026-03-20
missed deadlines outputting to CAN

### LINKED ISSUE #525: Add introspection into missed deadlines for RTOS on microcontrollers
It is possible that we miss executions for our RT threads. We should
1. Detect when a deadline is missed
2. Output these deadlines onto CAN

Do this for the 1, 10, 100, and 1k Hz threads in addition to each CAN message period

Nurtrantpoem19:
hey josh, would i need to create my own pack function to output missed deadlines? would that mean that i need to generate a can message in Messagegenerated.h?

### DIFF
--- a/network/definition/data/components/bmsb/bmsb-message.yaml
+++ b/network/definition/data/components/bmsb/bmsb-message.yaml
@@ -158,3 +158,11 @@ messages:
     sourceBuses: veh
     signals:
       packPower:
+  
+  missedDeadline:
+    description: BMS Boss missed deadline count
+    cycleTimeMs: 10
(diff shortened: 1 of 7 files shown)


network/definition/data/components/bmsb/bmsb-message.yaml
Nurtrantpoem19, line 1: @JoshLafleur Hey Josh, am i suppose to add this message to every single component in network/data/component? is this the right template for it? I'm not really sure what to do for Id and cycle time since this variable is suppose to track all the different missed deadliens for all the different periodicities.

PR #551, opened 2026-03-21
BMSW and BMSB Refactor

joshlafleur-zip:
### Reason for Change

We used to have the SOC and current limit calculations done on the BMSW and communicated to the boss. Move this decision making onto the BMSB for CFR26 to support an improved SOC estimation algorithm based on total pack characteristics.

Subsequently, we need to clean up lots of the fault handling behavior in both the BMSW and BMSB. We also want to support the new fault manager application to better identify failures in logs.

### Changes

1. BMSW
2. BMSB

### Test Plan

- [ ] Ensure the pack can be charged
- [ ] Ensure the pack can be discharged
- [ ] Probably more than this but we dont even have a pack yet :/

### DIFF
--- a/components/bms_boss/src/BMS.c
+++ b/components/bms_boss/src/BMS.c
@@ -31,7 +31,8 @@
  ******************************************************************************/
 
 #define CURRENT_SENSE_V_per_A -0.0025f
-#define PRECHARGE_MIN_TIME_MS 1320U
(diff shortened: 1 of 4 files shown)


joshlafleur-zip:
I _should_ do more testing here, but this world _should_ also be a better place so lgtm!

PR #560, opened 2026-04-03
bms_worker: NX3_NEN low on init for variant 0

KaileeSantucci:
### Reason for Change

`HW_GPIO_NX3_NEN` is configured with `resetState = HW_GPIO_PINSET` in the GPIO pinmux, which drives the pin high on init via `HW_GPIO_init()`. Since the pin is active low, the NX3 was never being enabled on variant 0 hardware, and no driver was bringing it low after init. It needs to be explicitly driven low during ADC init.

### Changes

1. Assert `HW_GPIO_NX3_NEN` low in `HW_ADC_init_componentSpecific()` under the `APP_VARIANT_ID == 0U` block

### Test Plan

- [ ]  Flash variant 0 hardware and verify NX3 is enabled and functioning after init

### DIFF
--- a/components/bms_worker/HW/HW_adc_componentSpecific.c
+++ b/components/bms_worker/HW/HW_adc_componentSpecific.c
@@ -133,6 +133,7 @@ HW_StatusTypeDef_E HW_ADC_init_componentSpecific(void)
     {
         Error_Handler();
     }
+    HW_GPIO_writePin(HW_GPIO_NX3_NEN, false);
 #elif APP_VARIANT_ID == 1U
     sConfig.Channel      = ADC_CHANNEL_TEMP_BOARD;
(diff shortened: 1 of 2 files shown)


components/bms_worker/HW/HW_adc_componentSpecific.c
```diff
@@ -133,6 +133,7 @@ HW_StatusTypeDef_E HW_ADC_init_componentSpecific(void)
     {
         Error_Handler();
     }
+    HW_GPIO_writePin(HW_GPIO_NX3_NEN, false);
```
JoshLafleur: This should be in environment.c in the init function

JoshLafleur (approved):
damn super shmint

PR #562, opened 2026-04-05
NVM: Static assertions on struct sizes

### LINKED ISSUE #550: nvm: Statically assert nvm struct sizes
Currently throughout the repo the nvm struct sizes are not restricted. This allows a developper to change the size of a V1 struct without a compilation failure forcing the developer to make subsequent changes (adding a new version, support version updates, etc..). Correct this by Statically asserting a size for all nvm entries

KaileeSantucci:
### Reason for Change

NVM structs lacked static size assertions, meaning a developer could modify a struct's layout without a compilation failure. Static assertions have been added to all NVM structs to enforce deterministic sizes at compile time.

### Changes

1. Added `_Static_assert(sizeof(...))` for all NVM structs across the repo to enforce deterministic sizes at compile time

### Test Plan

- [ ]  Build passes for all targets (`cfr25-embedded`, `cfr26-embedded`) with no static assertion failures

### DIFF
--- a/components/bms_boss/include/BMS.h
+++ b/components/bms_boss/include/BMS.h
@@ -104,6 +104,9 @@ extern nvm_bmsbContactorData_S contactor_data;
 
 extern BMSB_S BMS;
 
+NVM_SIZE_ASSERT(nvm_bmsData_S, (4U + (4U * BMS_CONFIGURED_SERIES_SEGMENTS * BMS_CONFIGURED_SERIES_CELLS) + 16U));
+NVM_SIZE_ASSERT(nvm_bmsbContactorData_S, 28U);
+
 /******************************************************************************
  *                       P U B L I C  F U N C T I O N S
  ******************************************************************************/
--- a/components/bms_boss/include/BMS.h
+++ b/components/bms_boss/include/BMS.h
@@ -104,6 +104,8 @@ extern nvm_bmsbContactorData_S contactor_data;
 
 extern BMSB_S BMS;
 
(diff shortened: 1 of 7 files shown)


components/shared/code/app/app_vehicleSpeed.h
```diff
@@ -60,6 +60,10 @@ typedef struct
 extern nvm_odometer_S odometer_data;
 #endif
 
+#if FEATURE_IS_ENABLED(FEATURE_VEHICLESPEED_LEADER)
+_Static_assert(sizeof(nvm_odometer_S) == 24U, "NVM deterministic size");
```
JoshLafleur: Were using this pattern so much it seems worthwhile to define something in the nvm header. Lik
```
#define NVM_SIZE_ASSERT(entry, size)
```

components/bms_boss/include/BMS.h
```diff
@@ -104,6 +104,9 @@ extern nvm_bmsbContactorData_S contactor_data;
 
 extern BMSB_S BMS;
 
+NVM_SIZE_ASSERT(nvm_bmsData_S, (4U + (4U * BMS_CONFIGURED_SERIES_SEGMENTS * BMS_CONFIGURED_SERIES_CELLS) + 16U));
```
JoshLafleur: We should actually define the number here, not use macros to prevent unintended unvalidated regressions

JoshLafleur (approved):
lgtm!

PR #564, opened 2026-04-05
bmsw: Add debug thermistor voltage CAN signals

KaileeSantucci:
### Reason for Change

### Changes

### Test Plan

- \[ \] Describe tests needed to validate the software changes


JoshLafleur:
Closing, brought into a seperately merged branch

PR #591, opened 2026-04-25
Improve vehicle speed calculation

JoshLafleur:
### Reason for Change

Identify locked/degraded wheel speed sensors. Integrate the IMU data to get an improved vehicle speed understanding, then update the vehicle speed with GPS and wheel speed measurements. Support the vehicle speed in vcpdu so it can inform behaviors based on vehicle speed.

### Changes

### Test Plan


JoshLafleur:
lgtm

Issue #219, opened 2025-04-14
Add current sense driver


JoshLafleur:

CodenameUsama:
I’m currently working on this. Feel free to reach out to me on Teams (Usama Aldakak) if needed.

PR #603, opened 2026-05-09
Add generic current sensing driver

CodenameUsama:
### Reason for Change

- Improve software uniformity by introducing a generic current sensing interface.
- Allow current output conversion through a shared built-in function.

### Changes

- Added a generic current sensing driver
- Reused current conversion logic in VN9008, TPS2HB16AB, and PDU/BMS current sensing paths
- Added current sensing source files to relevant BUCK targets

### Test Plan

- N/A


components/bms_boss/src/BMS.c
```diff
@@ -415,7 +413,9 @@ static void BMS1kHz_PRD(void)
     const uint64_t this_step = HW_TIM_getBaseTick();
     const uint32_t delta_t = (uint32_t)(this_step - BMS.counted_coulombs.last_step_us);
 
-    float32_t tmpCurrent =  (drv_inputAD_getAnalogVoltage(DRV_INPUTAD_ANALOG_CS) - PACK_CS_0_OFFSET) / CURRENT_SENSE_V_per_A;
+    float32_t tmpCurrent =  drv_currentSense_voltageToCurrent(
+    drv_inputAD_getAnalogVoltage(DRV_INPUTAD_ANALOG_CS) - PACK_CS_0_OFFSET,
+    1.0f / CURRENT_SENSE_V_per_A);
```
JoshLafleur, line 418: Checkout https://github.com/concordia-fsae/firmware/blob/master/components/shared/code/DRV/drv_tempSensors.h as a reference for this. Ideally we would define a set of component specific channels like https://github.com/concordia-fsae/firmware/blob/5f0e753924243e20b3c5f9cec43428639e82e057/components/bms_worker/HW/drv_tempSensors_componentSpecific.c#L2 and then we would get the current by calling `drv_currentSense_getCurrent(DRV_CURRENTSENSE_PACK_VOLTAGE)`.

What you have here would work but youre paying an additional complexity cost with no gain (all the driver does is a multiplication)

JoshLafleur (commented):
Please fix these items and clean up your test plan+commits. I can confirm the builds are not broken as per the latest PR into master https://github.com/concordia-fsae/firmware/actions/runs/25593360800, maybe try to rebase?

components/bms_boss/src/BMS.c
CodenameUsama: I was intentionally keeping this implementation lightweight for now since the main goal was to create a generic helper for current sensing across multiple modules. I agree that a full channel-based abstraction similar to the temp sensor driver could make sense if we start having more current sensing use cases. If we move in that direction later on, I’d be happy to jump back on it and help refactor it accordingly.

CodenameUsama:
related to #219

PR #607, opened 2026-05-12
Brake Temp sensors

freeman803:
INFKL-800 brake sensor

Reason for Change
 
Allows measuring rotor temperature dynamically 

Changes
1. Added brakeTemp.c and brakeTemp.h to components/vc/shared/app/
 
Test Plan
 -flash vcfront and vcrear.
- Confirm brakeTempFL and brakeTempFR CAN signals are present and updating on the bus at 10Hz
- ensure it reads correct room temperature


JoshLafleur (approved):
Good enough for this school

PR #617, opened 2026-05-29
updated battery model current limiting

davi4g1:
### Reason for Change

Current limiting formula of battery model is more accurate. The result is now added onto the existing pack current.

### DIFF
--- a/components/bms_boss/include/CANIO_componentSpecific.h
+++ b/components/bms_boss/include/CANIO_componentSpecific.h
@@ -46,6 +46,8 @@ CAN_prechargeContactorState_E CANIO_tx_getContactorState(void);
 
 #define set_packChargeLimit(m, b, n, s)                set(m, b, n, s, BMS.charge_limit)
 #define set_packDischargeLimit(m, b, n, s)             set(m, b, n, s, BMS.discharge_limit)
+#define set_maxDischargeModeled(m, b, n, s)            set(m, b, n, s, bm.dischargeLimit)
+#define set_maxChargeModeled(m, b, n, s)               set(m, b, n, s, bm.chargeLimit)
 #define set_packVoltage(m, b, n, s)                    set(m, b, n, s, BMS_VPACK_SOURCE)
 #define set_connectedSegments(m, b, n, s)              set(m, b, n, s, BMS.connected_segments)
 #define set_packVoltageCalculated(m, b, n, s)          set(m, b, n, s, BMS.pack_voltage_calculated)
--- a/components/bms_boss/include/CANIO_componentSpecific.h
+++ b/components/bms_boss/include/CANIO_componentSpecific.h
@@ -46,7 +46,7 @@ CAN_prechargeContactorState_E CANIO_tx_getContactorState(void);
 
 #define set_packChargeLimit(m, b, n, s)                set(m, b, n, s, BMS.charge_limit)
 #define set_packDischargeLimit(m, b, n, s)             set(m, b, n, s, BMS.discharge_limit)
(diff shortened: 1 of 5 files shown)


network/definition/data/components/bmsb/bmsb-signals.yaml
```diff
@@ -206,6 +206,32 @@ signals:
...
+  maxChargeModeled:
+    unit: 'A'
+    description: BMS Battery Model Charge Limit Raw
+    nativeRepresentation:
+      endianness: little
+      bitWidth: 10
+      resolution: 0.1
+      range:
+        min: 0
+        max: 75
```
JoshLafleur, line 250: Is this enough max? 10bit .1 res can do 0-100

components/bms_boss/include/CANIO_componentSpecific.h
```diff
@@ -46,6 +46,8 @@ CAN_prechargeContactorState_E CANIO_tx_getContactorState(void);
 
 #define set_packChargeLimit(m, b, n, s)                set(m, b, n, s, BMS.charge_limit)
 #define set_packDischargeLimit(m, b, n, s)             set(m, b, n, s, BMS.discharge_limit)
+#define set_maxDischargeModeled(m, b, n, s)            set(m, b, n, s, bm.dischargeLimit)
```
JoshLafleur: This needs to be *-1

components/bms_boss/src/batteryModel.c
```diff
@@ -119,16 +119,16 @@ static void model_state_run(batteryModel_S* batteryModel, float32_t cellVoltage,
...
 
-    RiDischarge                  = RiDischarge * 1.2f; // safety factor find maximum deviation from average in cell testing
-    RiCharge                     = RiCharge * 1.2f;    // safety factor find maximum deviation from average in cell testing
+    RiDischarge                  = RiDischarge * 1.0f; // safety factor find maximum deviation from average in cell testing
+    RiCharge                     = RiCharge * 1.0f;    // safety factor find maximum deviation from average in cell testing
 
-    batteryModel->dischargeLimit = (batteryModel->config.minCellVoltage - minCellVoltage) / RiDischarge;
-    batteryModel->chargeLimit    = (batteryModel->config.maxCellVoltage - maxCellVoltage) / RiCharge;
+    batteryModel->dischargeLimit = (batteryModel->config.minCellVoltage - minCellVoltage) / RiDischarge + cellCurrent;
+    batteryModel->chargeLimit    = (batteryModel->config.maxCellVoltage - maxCellVoltage) / RiCharge + cellCurrent;
```
JoshLafleur, line 137: I just thought of it but i think this only works when the current is the same direction? Ie if were discharging at 50A, we can discharge for battery model 50A, but for the charge case 50A discharge current shoudnt mean we can charge 50A less. I think we add the current to discharge limit if the pack is discharging, and same for the charge case

network/definition/data/components/bmsb/bmsb-signals.yaml
davi4g1: max is 75?

components/bms_boss/src/batteryModel.c
davi4g1: The math works out if you see it this way:
IR = 0.1V, min = 2.4V, live = 2.5V, cell charging
Model will say discharge limit = 0 because when you stop charging the IR drop will immediately put the cell at 2.4V because 2.5 - IR = 2.4.

JoshLafleur:
Closing and merging in https://github.com/concordia-fsae/firmware/pull/626

PR #635, opened 2026-06-05
added drive time remaining

davi4g1:
### Reason for Change

added time remaining for driving

Note: The end condition for this estimation is that drive time exceeds 30min. Since every step is 10 seconds, there is a maximum of 181 iterations. 

Also, the method model_state_run() was replaced with 2 methods:
model_states_run()
model_prediction_run()
so that the model states can be updated without the kalman filter. The kalman was placed in it's own function model_prediction_run(). This was done so that the drive time estimation can be implemented cleanly.

### DIFF
--- a/components/bms_boss/src/batteryModel.c
+++ b/components/bms_boss/src/batteryModel.c
@@ -20,7 +20,10 @@
  *                              D E F I N E S
  ******************************************************************************/
 
-#define  RI_SAFETY_FACTOR    1.0f;
+#define  RI_SAFETY_FACTOR              1.0f
+#define  DRIVETIME_AVG_WINDOW_S        60
+#define  DRIVETIME_MIN_SOC             0.01
+#define  DRIVETIME_MODEL_TIMESTEP_S    10
 
 /******************************************************************************
  *                             T Y P E D E F S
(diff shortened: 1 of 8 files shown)


network/definition/data/components/bmsb/bmsb-signals.yaml
```diff
@@ -251,6 +251,19 @@ signals:
       signedness: unsigned
     continuous: true
 
+  driveTimeRemaining:
+    unit: "m"
```
JoshLafleur: m is meter, you should add a min unit for time

network/definition/data/components/bmsb/bmsb-signals.yaml
```diff
@@ -251,6 +251,19 @@ signals:
...
+    description: BMS Battery Time Remaining to Drive
+    nativeRepresentation:
+      endianness: little
+      bitWidth: 12
+      resolution: 0.05
+      range:
+        min: 0
+        max: 180
```
JoshLafleur: I would say we have a 30 min upper bound no? 3hrs is pretty long

components/bms_boss/src/batteryModel.c
```diff
@@ -20,7 +20,10 @@
  *                              D E F I N E S
  ******************************************************************************/
 
-#define  RI_SAFETY_FACTOR    1.0f;
+#define  RI_SAFETY_FACTOR            1.0f
+#define  DRIVETIME_AVG_WINDOW        60
+#define  DRIVETIME_MIN_SOC           0.01
+#define  DRIVETIME_MODEL_TIMESTEP    10
```
JoshLafleur:
```suggestion
#define  DRIVETIME_MODEL_TIMESTEP_S  10
```

components/bms_boss/src/batteryModel.c
```diff
@@ -146,6 +159,22 @@ static void current_limit(batteryModel_S* batteryModel, float32_t minCellVoltage
...
+static void driveTime_remaining(batteryModel_S* batteryModel, float32_t cellVoltage, float32_t avgPower)
+{
+    int            numIterations   = 0;
+
+    batteryModel->driveTimeRemaining = 0;
+    batteryModel_S batteryModelTmp = *batteryModel;
+    batteryModelTmp.cellVoltageSim   = cellVoltage;
+    while (batteryModel_getSOC(&batteryModelTmp) > DRIVETIME_MIN_SOC && batteryModel->driveTimeRemaining <= 30 * 60)
+    {
+        // if (avgPower/cellVoltage < )
```
JoshLafleur: rm

components/bms_boss/src/batteryModel.c
```diff
@@ -146,6 +159,21 @@ static void current_limit(batteryModel_S* batteryModel, float32_t minCellVoltage
...
 }
 
+static void driveTime_remaining(batteryModel_S* batteryModel, float32_t cellVoltage, float32_t avgPower)
+{
+    int            numIterations   = 0;
+
+    batteryModel->driveTimeRemaining = 0;
+    batteryModel_S batteryModelTmp = *batteryModel;
+    batteryModelTmp.cellVoltageSim   = cellVoltage;
+    while (batteryModel_getSOC(&batteryModelTmp) > DRIVETIME_MIN_SOC && batteryModel->driveTimeRemaining <= 30 * 60)
```
JoshLafleur, line 169:
```suggestion
    while ((batteryModel_getSOC(&batteryModelTmp) > DRIVETIME_MIN_SOC) && (batteryModel->driveTimeRemaining <= (30 * 60)))
```
Its always more clear to be explicit in your order of operations

JoshLafleur (approved):

PR #641, opened 2026-06-12
edited cooling thresholds and added logic for overheated drivetrain

freeman803:
### Reason for Change

### Changes

### Test Plan

### DIFF
--- a/components/vc/pdu/src/cooling.c
+++ b/components/vc/pdu/src/cooling.c
@@ -20,12 +20,14 @@
 
 #define START_TIMER_MS                500U
 #define START_DUTY                    1.0f
-#define FAN_ON_DUTY                   0.7f
+#define FAN_ON_DUTY                   0.8f
 #define PUMP_ON_DUTY                  1.0f
 #define TEST_DUTY                     1.0f
+#define FAN_ON_DUTY_HOT               1.0f
 
-#define COOLING_LATCH_START_THRESH    50.0f
-#define COOLING_LATCH_STOP_THRESH     45.0f
+#define COOLING_LATCH_START_THRESH    40.0f
+#define COOLING_LATCH_STOP_THRESH     35.0f
+#define HOT_DRIVETRAIN                60.0f
 
 /******************************************************************************
  *                         P R I V A T E  V A R S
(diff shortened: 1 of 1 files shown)


components/vc/pdu/src/cooling.c
```diff
@@ -20,12 +20,14 @@
...
-#define FAN_ON_DUTY                   0.7f
+#define FAN_ON_DUTY                   0.8f
 #define PUMP_ON_DUTY                  1.0f
 #define TEST_DUTY                     1.0f
+#define FAN_ON_DUTY_HOT               1.0f
 
-#define COOLING_LATCH_START_THRESH    50.0f
-#define COOLING_LATCH_STOP_THRESH     45.0f
+#define COOLING_LATCH_START_THRESH    40.0f
+#define COOLING_LATCH_STOP_THRESH     35.0f
```
JoshLafleur: I think this is too low

components/vc/pdu/src/cooling.c
```diff
@@ -20,12 +20,14 @@
...
+#define HOT_DRIVETRAIN                60.0f
```
JoshLafleur: We probably wanna start this at 60 but turn it off at 55 and have some hysterisis, or add a ramp between the two

components/vc/pdu/src/cooling.c
```diff
@@ -131,10 +136,19 @@ static void cooling10Hz_PRD(void)
...
 
-    const float32_t dutyFan    = test.isTestFan ? TEST_DUTY : FAN_ON_DUTY;
-    const float32_t dutyPump   = test.isTestPump ? TEST_DUTY : PUMP_ON_DUTY;
-    const bool      enablePump = (isHV || isRun || test.isTestPump || cooling.drivetrainCoolingLatched) && !faultPump;
-    const bool      enableFan  = (isRun || test.isTestFan || cooling.drivetrainCoolingLatched) && !faultFan;
+    if (motorTemp > HOT_DRIVETRAIN_START_THRESH)
+    {
+        cooling.drivetrainHot = true;
+    }
+    if (motorTemp < HOT_DRIVETRAIN_STOP_THRESH)
```
JoshLafleur, line 143: Should technically be an else but its fine

JoshLafleur (approved):

PR #642, opened 2026-06-29
User/af/carputer export

freeman803:
### Reason for Change

### Changes

### Test Plan


davi4g1 (commented):
This code is meant to parse the csv table that is exported by influx?

freeman803:
> This code is meant to parse the csv table that is exported by influx?

It generates a CSV like influx but is run in command line and doesnt have the same limits as influx. influx limits you to 100mb csv this can export as much as you want

